PR #4, opened 2026-02-21
setup: prompt OLLAMA_API_BASE before credentials and use it for Ollama connectivity/model fetch

LiemFrans:
### PR Description (Changes)

- Prompt `OLLAMA_API_BASE` immediately after provider selection when `ollama` is selected.
- Normalize Ollama base URL with `_normalize_ollama_api_base()`:
  - blank input falls back to `http://localhost:11434`
  - auto-adds `http://` when scheme is missing
  - trims trailing `/`.
- Use selected `OLLAMA_API_BASE` for Ollama connectivity checks in credentials step.
- Use selected `OLLAMA_API_BASE` to fetch Ollama models in “Fetching Available Models”.
- Keep model-selection flow unchanged, but now free-model list reflects models from the user-provided Ollama host.
- Persist `OLLAMA_API_BASE` to `~/.nadirclaw/.env` in `write_env_file()`.
- Added/used base-aware connectivity helper (`_check_ollama_connectivity_with_base`).
- Updated setup wizard orchestration in `run_setup_wizard()` to pass Ollama base through credential + fetch steps.

### User-visible behavior

- For Ollama setups, hostname/base is requested **before**:
  1. Credentials connectivity check  
  2. Available model fetching  
  3. Model selection  

This ensures fetched Ollama models come from the intended endpoint.

### DIFF
--- a/nadirclaw/setup.py
+++ b/nadirclaw/setup.py
@@ -787,8 +787,7 @@ def write_env_file(
         lines.append("")
 
     if ollama_api_base:
-        if not api_keys:
-            lines.append("# API Keys")
+        lines.append("# Ollama")
         lines.append(f"OLLAMA_API_BASE={_normalize_ollama_api_base(ollama_api_base)}")
         lines.append("")
 
--- a/tests/test_setup.py
+++ b/tests/test_setup.py
@@ -11,10 +11,12 @@ from click.testing import CliRunner
 
 from nadirclaw.setup import (
     ENV_FILE,
+    _check_ollama_connectivity_with_base,
     _filter_anthropic_top,
     _filter_google_top,
     _filter_openai_top,
     _filter_top_models,
+    _normalize_ollama_api_base,
     classify_model_tier,
     detect_existing_config,
     fetch_provider_models,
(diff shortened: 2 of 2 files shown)


tests/test_setup.py
```diff
@@ -112,6 +114,71 @@ def test_gemini_pro_is_complex(self):
...
+
+    def test_custom_host_and_port(self):
+        assert _normalize_ollama_api_base("http://192.168.1.10:11434") == "http://192.168.1.10:11434"
+
+
+# ---------------------------------------------------------------------------
+# _check_ollama_connectivity_with_base
+# ---------------------------------------------------------------------------
+
+class TestCheckOllamaConnectivityWithBase:
```
doramirdor: doesn't cover the empty string case — _check_ollama_connectivity_with_base(""). This is the most likely "user pressed Enter without input" scenario and should fall back to the default, but it's not explicitly tested.

nadirclaw/setup.py
```diff
@@ -758,6 +786,11 @@ def write_env_file(
             lines.append(f"{env_var}={value}")
         lines.append("")
 
+    if ollama_api_base:
+        lines.append("# Ollama")
+        lines.append(f"OLLAMA_API_BASE={_normalize_ollama_api_base(ollama_api_base)}")
```
doramirdor: By the time write_env_file is called from run_setup_wizard, ollama_api_base is already normalized. The redundant _normalize_ollama_api_base() call inside write_env_file creates a hidden dependency on idempotency. Normalization should be the caller's responsibility; write_env_file should treat its parameter as already clean.

doramirdor:
Thanks for this contribution! I've opened #17 which includes your setup wizard changes plus the server-side fix I flagged in review (passing api_base to LiteLLM at request time). Going to close this in favor of that PR.

Issue #1, opened 2026-02-12
logo


doramirdor:
<img width="1536" height="1024" alt="Image" src="https://github.com/user-attachments/assets/80a5d910-fa91-4a36-80ad-6a59e2d8a14a" />

PR #8, opened 2026-02-22
feat: add tool calling support for LiteLLM providers

huveewomg:
## Summary
  - Forward `tools` and `tool_choice` from incoming requests to LiteLLM
  - Preserve `tool_calls`, `tool_call_id`, and `name` fields in conversation messages
  - Return `tool_calls` from LLM responses in both JSON and SSE streaming formats
  - Auto-upgrade `ollama/` to `ollama_chat/` when tools are present (LiteLLM's `ollama/` uses `/api/generate` which doesn't support tool calling)

## Why this is needed
Previously, tool definitions sent by clients were silently dropped in `_call_litellm()` because messages were flattened to `{"role", "content"}` only. This meant Ollama (and other LiteLLM providers) never received tool definitions, so tool calling always failed through NadirClaw even though it worked when calling the provider directly

## Test plan
  - [x] 17 new tests covering all tool calling flows
  - [x] Verified locally with Ollama + qwen3:4b — tool_calls returned correctly

## Note:
`test_server.py` and `test_tool_calling.py` both define a local `client` fixture. This is intentional duplication to avoid touching unrelated files in this PR A shared `conftest.py` would be the better long-term approach.

### DIFF
--- a/tests/test_tool_calling.py
+++ b/tests/test_tool_calling.py
@@ -227,10 +227,11 @@ class TestCallLitellmMessages:
         call_kwargs = mock_comp.call_args[1]
         messages = call_kwargs["messages"]
 
-        # Assistant message should have tool_calls
+        # Assistant message should have tool_calls and content: None (not "")
         assistant_msg = messages[1]
         assert "tool_calls" in assistant_msg
         assert assistant_msg["tool_calls"] == [SAMPLE_TOOL_CALL]
+        assert assistant_msg["content"] is None
 
         # Tool message should have tool_call_id and name
         tool_msg = messages[2]
@@ -261,6 +262,11 @@ class TestCallLitellmMessages:
         assert "tool_calls" in result
         assert result["tool_calls"] == [SAMPLE_TOOL_CALL]
         assert result["finish_reason"] == "tool_calls"
+
+        # Verify tool_calls round-trips through JSON serialization without TypeError
+        serialized = json.dumps(result)
+        deserialized = json.loads(serialized)
+        assert deserialized["tool_calls"] == [SAMPLE_TOOL_CALL]
 
     @pytest.mark.asyncio
     async def test_no_tool_calls_in_response_when_absent(self):
@@ -369,20 +375,20 @@ class TestNonStreamingToolCalls:
 class TestStreamingToolCalls:
     """Verify tool_calls appear in SSE stream chunks."""
 
-    @pytest.mark.parametrize("response_data,expected_key,expected_value", [
+    @pytest.mark.parametrize("response_data,expected_key,expected_value,expected_finish", [
         (
             {"content": None, "finish_reason": "tool_calls", "prompt_tokens": 10,
              "completion_tokens": 5, "tool_calls": [SAMPLE_TOOL_CALL]},
-            "tool_calls", [SAMPLE_TOOL_CALL],
+            "tool_calls", [SAMPLE_TOOL_CALL], "tool_calls",
         ),
         (
             {"content": "Hello world", "finish_reason": "stop",
              "prompt_tokens": 10, "completion_tokens": 5},
-            "content", "Hello world",
+            "content", "Hello world", "stop",
         ),
     ])
-    def test_streaming_delta(self, response_data, expected_key, expected_value):
-        """SSE stream delta should contain the expected key/value."""
+    def test_streaming_delta(self, response_data, expected_key, expected_value, expected_finish):
+        """SSE stream delta should contain the expected key/value and finish_reason."""
 
         sse_response = _build_streaming_response(
             request_id="test-123",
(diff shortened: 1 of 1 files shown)


doramirdor (commented):
## PR Review: Tool Calling Support for LiteLLM Providers

Nice work — this is a well-scoped feature that fills an important gap. The test coverage is solid (17 tests!) and the ollama→ollama_chat upgrade is a thoughtful detail. A few issues to address before merging:

---

### Critical

**1. `content: ""` vs `content: null` for tool-call assistant messages**

The new message-building loop uses `message.text_content()` which returns `""` when `content is None`. But the OpenAI/Anthropic APIs require `content: null` (not `""`) on assistant messages that carry `tool_calls`. Sending `content: ""` will cause API rejections from production providers.

Fix: use `message.content` directly instead of `message.text_content()`:
```python
msg: dict[str, Any] = {"role": message.role, "content": message.content}
```

Or conditionally:
```python
msg: dict[str, Any] = {"role": message.role}
content = message.text_content()
if content:
    msg["content"] = content
else:
    msg["content"] = message.content  # preserves None
```

**2. Pydantic tool_call objects may not JSON-serialize**

The response handling does:
```python
tool_calls = getattr(msg, "tool_calls", None)
if tool_calls:
    result["tool_calls"] = [
        tc.model_dump() if hasattr(tc, "model_dump") else tc
        for tc in tool_calls
    ]
```

This looks correct on paper — `model_dump()` should convert Pydantic objects to dicts. But worth adding a test that verifies the final JSON response actually round-trips through `json.dumps` without `TypeError`. The mock in `test_tool_calls_in_response` uses `MagicMock` with `model_dump` which is fine, but doesn't catch a real LiteLLM `ChatCompletionMessageToolCall` that might behave differently.

---

### Important

**3. SSE streaming puts `tool_calls` in the content chunk — protocol mismatch**

The OpenAI streaming protocol sends `tool_calls` as incremental deltas across multiple chunks. The current implementation sends the full `tool_calls` array in the first delta alongside `content`. While many clients tolerate this, strictly compliant parsers (especially the OpenAI SDK's streaming accumulator) expect the incremental format.

At minimum, when `tool_calls` is present, the first chunk's delta should have `content: null` rather than `content: ""` or the text content.

**4. `finish_reason` propagation**

The existing `or "stop"` fallback in `_call_litellm`:
```python
"finish_reason": response.choices[0].finish_reason or "stop",
```
This works today because `"tool_calls"` is truthy, but it's worth noting that some providers return `"tool_calls"` and others return `"function_call"` (legacy). As long as LiteLLM normalizes this, it's fine, but worth being aware of.

**5. Gemini fallback path doesn't handle tool calls**

If a tool-calling request rate-limits the primary model and falls back to a Gemini model via `_call_gemini`, the tool definitions won't be forwarded (since `_call_gemini` doesn't have tool-passing logic), and any `tool_calls` in the Gemini response won't be extracted. This means tool calling silently degrades on fallback. Consider either:
- Forwarding tools through `_call_gemini` too, or
- Logging a warning when tools are present but the fallback doesn't support them

---

### Minor / Nits

**6. Test file duplicates `client` fixture**

The PR notes this is intentional. A shared `conftest.py` would indeed be better — perhaps a fast follow-up?

**7. Test gap: `content: None` assertion**

Consider adding an assertion in `test_tool_calls_in_assistant_message_preserved` that the assistant message dict has `"content": None` (not `"content": ""`). This would catch Issue #1 directly.

**8. Test gap: `finish_reason: "tool_calls"` assertion**

`test_tool_calls_in_json_response` asserts `finish_reason == "tool_calls"` which is great. The streaming test should also verify the final chunk's `finish_reason`.

---

### Summary

| # | Issue | Severity |
|---|-------|----------|
| 1 | `content: ""` instead of `null` on tool-call messages | **Critical** — will fail against OpenAI/Anthropic |
| 2 | JSON serialization of tool_calls needs real-world validation | Medium |
| 3 | SSE streaming protocol mismatch for tool_calls | Medium |
| 4 | `finish_reason` propagation is fragile | Low |
| 5 | Gemini fallback silently drops tool calling | Medium |

**Verdict**: Please fix issue #1 (the `content: None` vs `""` bug) before merging — it will break tool calling against production providers. The rest are improvements that could be addressed in follow-ups.

doramirdor (approved):
## Second Review — LGTM ✅

The two fix commits cleanly address the critical and important issues from the first review:

### Issue 1 (Critical) — `content: null` preservation: **Fixed** ✅
```python
text = message.text_content()
msg: dict[str, Any] = {"role": message.role, "content": text if text else message.content}
```
When `content=None`, `text_content()` returns `""` (falsy), so it correctly falls through to `message.content` which preserves `None`. New test asserts `assistant_msg["content"] is None`.

### Issue 2 — JSON serialization: **Test added** ✅
`json.dumps(result)` round-trip assertion in `test_tool_calls_in_response` catches any `TypeError` from un-serialized Pydantic objects.

### Issue 3 (Important) — SSE streaming protocol: **Fixed** ✅
```python
if tool_calls:
    delta["tool_calls"] = tool_calls
    delta["content"] = None
else:
    delta["content"] = content
```
Content is `None` (not `""`) when tool_calls are present. Test asserts both `delta["content"] is None` and correct `finish_reason` in the terminal chunk.

### Issues 4, 5, 6 — acknowledged as follow-ups, fine for this PR scope.

Clean work addressing the feedback. Approving.

PR #27, opened 2026-03-13
feat: upgrade-only session cache for agent frameworks

rmarji:
## Problem

When used with agent frameworks like OpenClaw, the session cache pins the routing tier based on the first message in a session. Since agent frameworks send large system prompts + tool schemas, the first classification often lands on "mid" — and the 30-minute TTL means every subsequent message (even genuinely complex ones) stays pinned to mid/sonnet.

This is the same class of issue that affects LiteLLM's `complexity_router` with tool-heavy payloads.

## Solution

**Upgrade-only session cache** — the cached tier can only escalate, never downgrade.

### Changes

- **Always classify** every request (no more cache-hit short-circuit)
- **New method `upgrade_if_higher()`** — atomically compares new tier against cached tier, upgrades if higher, keeps cached if equal or above
- **`put()` is upgrade-only** — skips write if cached tier already outranks the new one
- **TTL reduced** from 30min to 5min for faster session turnover
- **Tier hierarchy**: simple (0) < mid (1) < complex (2) < reasoning (3)

### Behavior

| Step | Classified | Cached | Used | |
|------|-----------|--------|------|-|
| 1 | simple | (none) | claw-quick | New session |
| 2 | complex | simple→complex | claw-reason | ✅ Upgraded |
| 3 | simple | complex | claw-reason | ✅ No downgrade |

### Tested with

- OpenClaw + NadirClaw + LiteLLM (3-tier: claw-quick / claw-chat / claw-reason)
- Verified: simple→quick, complex→reason, simple-after-complex→stays reason
- Streaming and non-streaming paths both work

### DIFF
--- a/nadirclaw/routing.py
+++ b/nadirclaw/routing.py
@@ -329,9 +329,16 @@ class SessionCache:
 
     Keyed by a hash of the system prompt + first user message.
     TTL-based expiry with LRU eviction to cap memory usage.
+
+    Upgrade-only policy: cached tier can only escalate (simple→mid→complex→
+    reasoning), never downgrade.  This prevents a complex session from being
+    pinned to "simple" while still avoiding jarring model switches downward.
     """
 
-    def __init__(self, ttl_seconds: int = 1800, max_size: int = 10_000):
+    # Tier ordering — higher index = more capable model.
+    TIER_ORDER = {"simple": 0, "mid": 1, "complex": 2, "reasoning": 3}
+
+    def __init__(self, ttl_seconds: int = 300, max_size: int = 10_000):
         # OrderedDict gives O(1) move-to-end (move_to_end) and O(1) popitem(last=False)
         # for LRU eviction — replaces the old List-based access_order which was O(n).
         self._cache: OrderedDict[str, Tuple[str, str, float]] = OrderedDict()  # key → (model, tier, timestamp)
@@ -372,7 +379,12 @@ class SessionCache:
             self._cache.popitem(last=False)
 
     def get(self, messages: List[Any]) -> Optional[Tuple[str, str]]:
-        """Return (model, tier) if a session exists and isn't expired."""
+        """Return (model, tier) if a session exists and isn't expired.
+
+        The caller is expected to *always* run the classifier after this.
+        If the new classification yields a higher tier, call
+        ``upgrade_if_higher`` to atomically escalate the cached entry.
+        """
         key = self._make_key(messages)
         with self._lock:
             entry = self._cache.get(key)
@@ -385,15 +397,53 @@ class SessionCache:
             self._touch(key)
             return model, tier
 
-    def put(self, messages: List[Any], model: str, tier: str) -> None:
-        """Store a routing decision for this session."""
+    def upgrade_if_higher(
+        self, messages: List[Any], new_model: str, new_tier: str
+    ) -> Tuple[str, str]:
+        """Upgrade the cached tier if *new_tier* outranks the stored one.
+
+        Returns the (model, tier) that should actually be used — either the
+        newly upgraded values or the previously cached ones.
+        """
         key = self._make_key(messages)
+        new_rank = self.TIER_ORDER.get(new_tier, 0)
+        with self._lock:
+            entry = self._cache.get(key)
+            if entry is None:
+                # Nothing cached yet — store and return the new values.
+                self._cache[key] = (new_model, new_tier, time.time())
+                return new_model, new_tier
+            cached_model, cached_tier, _ts = entry
+            cached_rank = self.TIER_ORDER.get(cached_tier, 0)
+            if new_rank > cached_rank:
+                # Escalate — upgrade the cache entry.
+                self._cache[key] = (new_model, new_tier, time.time())
+                self._touch(key)
+                return new_model, new_tier
+            # Keep the existing (equal or higher) tier.
+            self._touch(key)
+            return cached_model, cached_tier
+
+    def put(self, messages: List[Any], model: str, tier: str) -> None:
+        """Store a routing decision for this session (upgrade-only).
+
+        If an entry already exists with a higher tier, this is a no-op.
+        """
+        key = self._make_key(messages)
+        new_rank = self.TIER_ORDER.get(tier, 0)
         with self._lock:
             # Periodic cleanup of expired entries
             self._cleanup_counter += 1
             if self._cleanup_counter >= self._cleanup_interval:
                 self._cleanup_counter = 0
                 self.clear_expired()
+
+            # Upgrade-only: don't downgrade an existing entry.
+            existing = self._cache.get(key)
+            if existing is not None:
+                _, cached_tier, _ = existing
+                if self.TIER_ORDER.get(cached_tier, 0) >= new_rank:
+                    return  # existing tier is equal or higher — skip
 
             self._cache[key] = (model, tier, time.time())
             self._touch(key)
@@ -415,7 +465,7 @@ class SessionCache:
 
 
 # Global session cache
-_session_cache = SessionCache(ttl_seconds=1800)
+_session_cache = SessionCache(ttl_seconds=300)
 
 
 def get_session_cache() -> SessionCache:
--- a/nadirclaw/routing.py
+++ b/nadirclaw/routing.py
@@ -399,30 +399,41 @@ class SessionCache:
 
     def upgrade_if_higher(
         self, messages: List[Any], new_model: str, new_tier: str
-    ) -> Tuple[str, str]:
+    ) -> Tuple[str, str, str]:
         """Upgrade the cached tier if *new_tier* outranks the stored one.
 
-        Returns the (model, tier) that should actually be used — either the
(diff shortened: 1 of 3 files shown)


doramirdor (commented):
## Review

Good concept — upgrade-only session caching is the right fix for agent frameworks that inflate the first classification. The core invariant (tier can only escalate, never downgrade) is clean and easy to reason about. A few issues below before this is ready to merge.

### Summary

| Category | Item |
|----------|------|
| **Bug** | `upgrade_if_higher` doesn't check TTL — stale high-tier entries block fresh classifications |
| **Bug** | No `_evict_lru()` call in `upgrade_if_higher` — cache can grow past `max_size` |
| **Observability** | `analysis_info` no longer reflects whether the cache was used vs. fresh classification |
| **Testing** | No tests for `upgrade_if_higher()` or the new upgrade-only guard in `put()` |
| **Cleanup** | `put()` is effectively dead code in the server hot path |

doramirdor (changes requested):

doramirdor (changes requested):
## Code Review

### 🚨 Critical Issues

**1. TOCTOU race in `_init_feedback_tables` (feedback.py:43)**
- `_tables_initialized` is read outside `_db_lock`, set inside it. Two concurrent threads can both pass the guard. 
- **Fix:** Add double-checked locking:
```python
def _init_feedback_tables() -> None:
    global _tables_initialized
    if _tables_initialized:
        return
    with _db_lock:
        if _tables_initialized:   # ← add this re-check
            return
        # ... DDL ...
        _tables_initialized = True
```

**2. `get_rules_engine` singleton has no lock (rules.py:382)**
- Unlike `get_circuit_breaker()` which correctly uses a double-checked lock, `get_rules_engine()` has no lock at all. Two concurrent requests at startup can both construct and load the engine, racing on the global assignment.
- **Fix:** Follow the same pattern as `get_circuit_breaker()` — add a module-level `Lock` and double-checked locking.

### ⚠️ Important Issues

**3. All four new modules use deprecated `typing.Dict/List/Optional`**
- `CONTRIBUTING.md` explicitly requires Python 3.10+ syntax: `dict` not `Dict`, `list` not `List`, `X | None` not `Optional[X]`
- All four new files (`feedback.py`, `rules.py`, `optimizer.py`, `circuit_breaker.py`) use the old-style imports throughout.

**4. No hour/minute bounds validation in `_check_time_range` (rules.py:318)**
- `"25:99-26:00"` passes the split check and produces `start_minutes = 1599` — silently never matches instead of warning.

**5. `record_failure` in HALF_OPEN doesn't reset `consecutive_failures` (circuit_breaker.py:182)**
- When a probe fails in HALF_OPEN, the counter retains the pre-cooldown count. Logged failure counts become misleadingly inflated across OPEN → HALF_OPEN → OPEN cycles. Not a state-machine bug, but confusing for operators.

**6. No runtime reload mechanism for `rules.yaml`**
- `reload_rules()` exists but is not wired to any endpoint, signal handler, or file watcher. Operators must restart the server to pick up rule changes. Should either document this limitation or expose a `/v1/rules/reload` endpoint.

doramirdor:
@rmarji — friendly ping. This PR has been dormant since 2026-03-28. Two reviews on the latest commit (`5467a5c`) still show unaddressed blockers:

**From maintainer review (2026-03-14):**
1. `upgrade_if_higher` doesn't check TTL — stale high-tier entries block fresh classifications
2. No `_evict_lru()` call in `upgrade_if_higher` — cache can grow past `max_size`
3. `analysis_info` no longer reflects cache vs. fresh classification (observability regression)
4. No tests for `upgrade_if_higher()` or the upgrade-only guard in `put()`
5. `put()` is effectively dead code in the server hot path — candidate for removal

**From later review (2026-03-28):** TOCTOU race in `_init_feedback_tables`, missing lock on `get_rules_engine()` singleton, deprecated `typing.Dict/List/Optional` usage, `time_range` hour/minute bounds not validated. _(These look like they may be in a broader rebase-merge and not directly tied to this feature, but they block the CI story.)_

The core idea (upgrade-only tier escalation) is sound and solves a real problem for OpenClaw-style agent frameworks. If you're still interested in landing this, a rebase + fixes for (1)–(4) above should be enough to unblock. Otherwise, happy to close if it's been superseded.

---

doramirdor:
@rmarji — final notice. This PR has been dormant since 2026-03-14 (~6 weeks). The core idea (upgrade-only tier escalation) genuinely solves a problem for OpenClaw-style agent frameworks where a tool-heavy first message inflates the cached tier. Outstanding from the 2026-03-14 review on `5467a5c`:

1. **`upgrade_if_higher` doesn't check TTL** — stale high-tier entries block fresh classifications.
2. **No `_evict_lru()` call in `upgrade_if_higher`** — cache can grow past `max_size`.
3. **`analysis_info` no longer reflects cache vs. fresh classification** — observability regression.
4. **No tests for `upgrade_if_higher()` or the upgrade-only guard in `put()`**.

(The 2026-03-28 review mixed in cross-PR rebase noise that doesn't apply here — ignore that one; only the 2026-03-14 items are blocking.)

Plan:

- **By 2026-05-13** — if you can address (1)–(4), I'll re-review and merge.
- **After 2026-05-13** — I'll either close as stale or carry the feature forward with `Co-Authored-By` attribution to you. Let me know which you prefer.

Thanks for raising this — it's a real failure mode and worth landing.

---

rmarji:
Pushed [d5ccd12](https://github.com/rmarji/NadirClaw/commit/d5ccd12) addressing items 1–4:

**1. TTL check in `upgrade_if_higher`** — stale (>TTL) entries are now discarded before the upgrade check, so a high-tier entry that's outlived its window cannot block a fresh classification. ([routing.py:418-422](https://github.com/rmarji/NadirClaw/blob/feature/upgrade-only-session-cache/nadirclaw/routing.py#L418-L422))

**2. `_evict_lru()` in `upgrade_if_higher`** — called after inserting a new entry so the cache cannot grow past `max_size`. This is now the hot path since the server only calls `upgrade_if_higher` (not `put`). ([routing.py:425](https://github.com/rmarji/NadirClaw/blob/feature/upgrade-only-session-cache/nadirclaw/routing.py#L425))

**3. Cache vs. fresh classification observability** — `upgrade_if_higher` now returns `(model, tier, status)` where `status` is `"new"` / `"upgraded"` / `"kept"`. The server records this as `analysis_info["cache_status"]` and appends `"+session-cache"` to the strategy when cached values won. ([server.py:1056-1067](https://github.com/rmarji/NadirClaw/blob/feature/upgrade-only-session-cache/nadirclaw/server.py#L1056-L1067))

**4. Tests** — 11 new tests in `TestSessionCache` covering: status codes (`new`/`upgraded`/`kept`), the `simple→mid→complex→reasoning` hierarchy walking up and back down, expired-entry handling (injected stale timestamp), `max_size` enforcement, LRU touch ordering across `upgrade_if_higher` calls, and the upgrade-only guard in `put()` (no-downgrade, equal-tier no-op, upgrade path). ([tests/test_routing.py:301-417](https://github.com/rmarji/NadirClaw/blob/feature/upgrade-only-session-cache/tests/test_routing.py#L301-L417))

Local CI-equivalent run (`pytest tests/ --ignore=tests/test_server.py`): **389 passed**.

Note on item 5: kept `put()` since item 4 calls for testing its upgrade-only guard. Happy to drop it in a follow-up if you'd rather narrow the public surface.

doramirdor:
@rmarji — verified d5ccd12 against the four blockers and everything checks out:

1. **TTL check** ✅ — [routing.py:418-422](https://github.com/rmarji/NadirClaw/blob/feature/upgrade-only-session-cache/nadirclaw/routing.py#L418-L422) discards expired entries before the rank comparison, so a stale `reasoning` entry can no longer pin a session that's actually `simple` now.
2. **`_evict_lru()` on the new hot path** ✅ — called right after the fresh insert in `upgrade_if_higher`. `max_size` capacity test (`test_upgrade_if_higher_evicts_when_over_capacity`) and the LRU touch test both pin this down.
3. **Observability** ✅ — `analysis_info["cache_status"]` is now `"new" | "upgraded" | "kept"`, and the strategy gets `+session-cache` appended only when the cached values actually won. Cleanly distinguishes the three cases.
4. **Tests** ✅ — the 11 new cases cover the upgrade-only guard in `put()`, the full `simple→mid→complex→reasoning` walk both directions, expired-entry passthrough, capacity enforcement, and LRU touch ordering. Nice coverage.

One non-blocking note: TTL default dropped from 1800s to 300s in this commit. Reasonable for the agent-session use case (shorter coherent windows), but worth calling out in the PR description / changelog since it's a behavior change for anyone already using `SessionCache` directly.

On `put()` retention — agreed, leaving it in is fine. The upgrade-only guard there is a useful defense-in-depth even though the server doesn't call it on the hot path.

Approving and merging. Thanks for sticking with this — it's a real improvement for OpenClaw-style flows.

---

doramirdor (approved):
All four blockers from the 2026-03-14 review addressed in d5ccd12. No security concerns — pure cache-logic changes with good test coverage.

PR #29, opened 2026-03-24
feat(security): production security hardening — OWASP API Top 10

pradumna-gautam:
## Summary

Comprehensive security audit and hardening of the `/v1/chat/completions` endpoint for production launch at 5k-10k RPM, targeting OWASP API Security Top 10 compliance.

### Critical Fixes
- **CORS**: Replaced `allow_origins=["*"]` + `allow_credentials=True` with explicit origin allowlist — eliminates credential theft from malicious cross-origin sites
- **Auth**: Constant-time token comparison via `hmac.compare_digest` — prevents timing side-channel brute-force
- **Security headers**: `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, `Referrer-Policy`, `Cache-Control: no-store`, opt-in HSTS
- **Validation errors**: Sanitized — no longer leaks Pydantic internals to clients
- **System prompt logging**: API key pattern redaction + 500-char truncation + opt-in
- **Input validation**: Bounds on messages (500), max_tokens (100K), temperature (0-2), top_p (0-1), n (1-8)

### New Security Modules
- **prompt_guard.py**: 7 heuristic injection patterns, configurable log/warn/block with response header
- **pii_redactor.py**: Email/phone/SSN/credit card (Luhn) redaction, configurable none/log_only/redact
- **replay_guard.py**: Opt-in `X-Request-Nonce` deduplication with 300s TTL
- **request_signing.py**: HMAC-SHA256 for service-to-service integrity
- **credentials.py**: Fernet encryption at rest with auto-migration from plaintext

### Performance Fix
- Replaced `threading.Lock` blocking event loop with `ThreadPoolExecutor` for all logging I/O

## Test plan
- [ ] Verify CORS: `curl -H "Origin: https://evil.com"` returns no ACAO header
- [ ] Verify security headers present on `/v1/chat/completions` responses
- [ ] Verify validation errors return sanitized messages (no Pydantic internals)
- [ ] Test prompt injection patterns are detected and logged
- [ ] Test PII redaction with `NADIRCLAW_PII_REDACTION=redact`
- [ ] Test replay guard rejects duplicate `X-Request-Nonce`
- [ ] Load test at 1K RPM to verify no event loop blocking from logging

### DIFF
--- a/nadirclaw/auth.py
+++ b/nadirclaw/auth.py
@@ -5,6 +5,7 @@ Supports both Authorization: Bearer <token> and X-API-Key: <token>
 so any OpenAI-compatible client works out of the box.
 """
 
+import hmac
 import json
 import logging
 import os
@@ -102,14 +103,22 @@ async def validate_local_auth(
             detail="Missing auth token. Send Authorization: Bearer <token> or X-API-Key: <token>",
         )
 
-    user_data = _LOCAL_USERS.get(token)
-    if user_data is None:
+    # Constant-time token comparison to prevent timing side-channel attacks.
+    # Iterate all configured tokens so timing does not reveal which (or whether)
+    # a token exists in the dict.
+    matched_user_data = None
+    for candidate_token, candidate_data in _LOCAL_USERS.items():
+        if hmac.compare_digest(token.encode("utf-8"), candidate_token.encode("utf-8")):
+            matched_user_data = candidate_data
+            # Do NOT break — iterate all keys for constant-time behavior
+
+    if matched_user_data is None:
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail="Invalid auth token",
         )
 
-    return UserSession(user_data)
+    return UserSession(matched_user_data)
 
 
 def _default_user() -> Dict[str, Any]:
--- a/nadirclaw/credentials.py
+++ b/nadirclaw/credentials.py
@@ -6,6 +6,8 @@ Supports OAuth tokens with automatic refresh for all providers.
 OpenClaw integration is optional — NadirClaw works standalone.
 """
 
+import base64
+import hashlib
 import json
 import logging
 import os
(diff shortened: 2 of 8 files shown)


doramirdor (commented):
## Security Hardening Review

Thanks for the thorough work here. About half of this PR addresses genuine production gaps, and the rest ranges from nice-to-have to premature. Suggestion: **split this into two PRs** — ship the core hardening now, defer the new modules.

---

### ✅ Genuinely Needed (ship these)

| Fix | Why |
|-----|-----|
| **CORS wildcard fix** | `allow_origins=["*"]` + `allow_credentials=True` is a real vulnerability — any malicious site can make credentialed cross-origin requests |
| **Constant-time auth** | Plain dict lookup enables timing side-channel attacks on API keys |
| **Security headers** | Missing `Cache-Control: no-store`, `X-Content-Type-Options`, etc. matters at scale |
| **Input bounds validation** | No cap on `max_tokens` or message count = direct cost exposure |
| **Async logging** | Sync SQLite writes in async routes will cause latency spikes at 5-10k RPM |
| **Prompt truncation** | Full prompts in logs is a data retention risk |

### 🟡 Nice-to-Have (harmless, off by default)

- **PII redactor** — only covers non-streaming responses, so incomplete protection model
- **Prompt guard** — false-positive risk on legitimate prompts like "act as a SQL expert". Also uses `assert` for config validation which gets stripped with `python -O` — should be a `ValueError` instead

### 🔴 Premature for Current Architecture

- **Replay guard** — wrong threat model for an LLM proxy (identical prompts are expected). Also the 50k nonce cap silently fails at 5-10k RPM (300s window × 10k RPM = up to 3M requests, cap evicts nonces before TTL expires)
- **Request signing** — solves a service mesh problem that doesn't exist yet (single-process server)
- **Credential encryption** — uses raw SHA-256 instead of a proper KDF (PBKDF2/scrypt/Argon2), so weak `NADIRCLAW_CREDENTIAL_KEY` values produce weak keys

### Bugs to Fix

1. **`prompt_guard.py`** — Replace `assert _ACTION in ("log", "warn", "block")` with `if ... raise ValueError(...)`. Asserts are stripped with `python -O`, silently disabling the guard.
2. **`replay_guard.py`** — Document or fix the `max_size=50000` eviction math. At target RPM, nonces get evicted before their TTL, so replays of evicted nonces pass through.

---

**TL;DR:** The 6 core fixes in `server.py`, `auth.py`, and `request_logger.py` are real production-readiness gaps — approve those. The 5 new modules are well-written but solving future problems. Consider splitting.

doramirdor (changes requested):
## Code Review

This PR appears to be the parent/superset of PRs #30 and #31, which were split from it. It contains the same modules and the same critical issues found in those PRs.

### 🚨 Critical Issues (same as #30 and #31)

1. **Auth crash** — trailing comma in `auth.py` creates tuple instead of dict → `AttributeError` on every authenticated request
2. **`trained_model.pkl` in repo** — `joblib.load()` = arbitrary code execution; co-located SHA-256 sidecar is self-defeating
3. **Invalid CORS wildcards** — `"http://localhost:*"` not supported by Starlette → breaks admin dashboard (regression)
4. **TOCTOU race** in `_init_feedback_tables` — missing double-checked locking
5. **`"mid"` vs `"medium"` tier mismatch** — corrupts training pipeline

### Recommendation

If #30 and #31 are the intended split of this PR, consider **closing #29** to avoid confusion and merge conflicts. The reviewer comment from doramirdor suggesting a split appears to have been acted on.

If this PR is still active, all critical issues from the #30 and #31 reviews apply here as well and need to be addressed before merge.

doramirdor:
## Recommendation: Close this PR in favor of #30 + #31

Hey @pradumna-gautam — it looks like this PR was split into #30 (core hardening) and #31 (optional modules), which matches the reviewer feedback from @doramirdor. Having all three open creates confusion and will cause merge conflicts.

**Suggestion:** Close this PR and continue work on #30 + #31 as the canonical split.

Both #30 and #31 have detailed reviews posted — the most urgent items to address are:

1. **Auth crash in #30** — trailing comma on `matched_user_data = candidate_data,` turns it into a tuple → `AttributeError` on every authenticated request (one-character fix)
2. **CORS regression in #30** — `"http://localhost:*"` is not valid in Starlette's `CORSMiddleware` (exact match only) → breaks the admin dashboard
3. **Tier name mismatch in #31** — `"mid"` vs `"medium"` across `feedback.py` and `request_logger.py` silently corrupts training data

See the full reviews on [#30](https://github.com/NadirClaw/NadirClaw/pull/30) and [#31](https://github.com/NadirClaw/NadirClaw/pull/31) for the complete list.

PR #30, opened 2026-03-26
fix(security): core production hardening — CORS, auth, headers, validation

pradumna-gautam:
## Summary

Ship the 6 genuinely needed security fixes for production readiness at 5-10k RPM. Removes the 5 premature security modules (replay guard, request signing, credential encryption, prompt guard, PII redactor) that were solving future problems not present in the current single-process architecture.

**Core fixes retained:**
- **CORS**: Replaced `allow_origins=["*"]` + `allow_credentials=True` with explicit origin allowlist — eliminates credential theft from malicious cross-origin sites
- **Auth**: Constant-time token comparison via `hmac.compare_digest` — prevents timing side-channel brute-force
- **Security headers**: `X-Content-Type-Options: nosniff`, `X-Frame-Options: DENY`, `Cache-Control: no-store`, opt-in HSTS
- **Input bounds validation**: Caps on messages (500), max_tokens (100K), temperature (0-2), top_p (0-1), n (1-8) — prevents cost exposure
- **Async logging**: `ThreadPoolExecutor` replaces blocking SQLite writes in async routes — eliminates latency spikes at scale
- **Prompt truncation**: 500-char limit in SQLite logs + API key pattern redaction

**Premature modules removed:**
- `replay_guard.py` — wrong threat model (identical prompts expected for LLM proxy), nonce eviction bug at target RPM
- `request_signing.py` — service mesh solution for non-existent problem (single-process)
- `credentials.py` encryption — raw SHA-256 instead of proper KDF (PBKDF2/scrypt/Argon2)
- `prompt_guard.py` and `pii_redactor.py` — deferred to separate PR (#nice-to-have)

Also retained: sanitized validation error responses (no Pydantic internals leaked to clients).

## Test plan

- [ ] Verify CORS: `curl -H "Origin: https://evil.com"` returns no ACAO header
- [ ] Verify security headers present on `/v1/chat/completions` responses
- [ ] Verify validation errors return sanitized messages (no Pydantic internals)
- [ ] Verify input bounds reject messages >500, max_tokens >100K, etc.
- [ ] Load test at 1K RPM to verify no event loop blocking from async logging
- [ ] Run full test suite: `pytest tests/ -x` (383 tests passing)


doramirdor (changes requested):
## Code Review

### 🚨 Critical Issues

**1. Auth crash — trailing comma creates tuple (auth.py)**
```python
matched_user_data = candidate_data,   # ← trailing comma!
```
This assigns a 1-tuple `(candidate_data,)` instead of `candidate_data`. When `UserSession(matched_user_data)` calls `.get()` on a tuple → `AttributeError` on every authenticated request. The `is None` check passes (tuple is not None), so 401 is never raised either.
- **Fix:** Remove the trailing comma: `matched_user_data = candidate_data`

**2. `trained_model.pkl` shipped in repo — arbitrary code execution risk**
- `joblib.load()` uses `pickle` internally → arbitrary code execution on untrusted input
- The SHA-256 sidecar is co-located on disk, so an attacker who replaces the model can replace the hash too — self-defeating
- **Fix:** Add `*.pkl` to `.gitignore`. If model must ship, pin the hash as a constant in source code, not a sidecar file.

**3. CORS wildcard patterns are invalid — breaks admin dashboard**
```python
_cors_origins = ["http://localhost:*", "http://127.0.0.1:*", ...]
```
Starlette's `CORSMiddleware` does exact string matching — `"http://localhost:*"` will never match `"http://localhost:8084"`. This is a regression from the current `allow_origins=["*"]` and will break all cross-origin requests from the dashboard.
- **Fix:** Enumerate specific ports (`["http://localhost:8084", ...]`) or use `["*"]` for local-only deployments.

**4. Module-level `os.getenv()` breaks project Settings pattern**
- `request_signing.py`: `_HMAC_SECRET = os.getenv("NADIRCLAW_INTERNAL_HMAC_SECRET")` 
- `pii_redactor.py`: `_MODE = os.getenv("NADIRCLAW_PII_REDACTION", "none")`
- The project uses `Settings` with dynamic `@property` reads so `serve` CLI env overrides work. Module-level reads bypass this — env vars set by `serve` are silently ignored.

**5. TOCTOU race in `_init_feedback_tables` (feedback.py:43)**
- Same issue as PR #31 — missing double-checked locking.

### ⚠️ Important Issues

**6. `classify_with_system()` omits confidence escalation (trained_classifier.py:392)**
- `classify()` escalates tier when confidence < 0.70, but `classify_with_system()` hardcodes `confidence_escalated: False`. Since `analyze()` calls `classify_with_system()` when a system prompt is present (the common case), the escalation logic is effectively dead code in production.

**7. `reload_rules()` creates a window where `_engine is None` (rules.py:395)**
```python
_engine = None            # ← concurrent caller gets None here
return get_rules_engine() # ← rebuilds
```
- **Fix:** Build new engine first, then swap atomically: `new_engine = RoutingRulesEngine(); new_engine.load(...); _engine = new_engine`

**8. Nonce store capacity too small for stated RPM target (replay_guard.py)**
- 50k capacity with 300s window at 10k RPM = 49,800 nonces/window — any burst triggers eviction and allows replays.

**9. Silent plaintext fallback when encryption is requested (credentials.py)**
- If `NADIRCLAW_CREDENTIAL_KEY` is set but `cryptography` package is missing, credentials are stored in plaintext with only a warning. Should raise an error when the user has explicitly opted into encryption.

pradumna-gautam:
@claude @doramirdor 


---

Thanks for the review. Investigated all 9 issues — **8 are invalid** (reference files/code that don't exist in this repo), **1 was valid and is now fixed**.

| # | Claim | Status |
|---|---|---|
| 1 | Trailing comma in `auth.py` | **Invalid** — actual code is `matched_user_data = candidate_data` (no comma) |
| 2 | `trained_model.pkl` in repo | **Invalid** — no `.pkl` files exist anywhere in the repo |
| 3 | CORS wildcard patterns invalid | **Valid — fixed in 14d7ac2.** Switched from `allow_origins` with `"http://localhost:*"` to `allow_origin_regex` with `r"^https?://(localhost\|127\.0\.0\.1)(:\d+)?$"` |
| 4 | Module-level `os.getenv()` in `request_signing.py` / `pii_redactor.py` | **Invalid** — neither file exists on this branch |
| 5 | TOCTOU race in `feedback.py:43` | **Invalid** — `feedback.py` doesn't exist |
| 6 | `classify_with_system()` in `trained_classifier.py:392` | **Invalid** — file doesn't exist |
| 7 | `reload_rules()` in `rules.py:395` | **Invalid** — file doesn't exist |
| 8 | Nonce store in `replay_guard.py` | **Invalid** — file was removed from this branch |
| 9 | Plaintext fallback in `credentials.py` | **Invalid** — encryption was removed from this branch; no `NADIRCLAW_CREDENTIAL_KEY` exists |

The CORS fix now uses `allow_origin_regex` for the default localhost config, while production deployments with explicit `NADIRCLAW_CORS_ORIGINS` continue using exact origin matching.

doramirdor (changes requested):
## Code Review

### Good changes
- **CORS regex fix** — properly anchored (`^...$`), uses `allow_origin_regex` correctly. Nice fix for the Starlette wildcard bug.
- **Sanitized validation errors** — no Pydantic internals leaked to clients.
- **Security headers** — standard best practices, opt-in HSTS is the right call.
- **Bounds validation on ChatCompletionRequest** — reasonable limits.
- **System prompt redaction** — API key patterns stripped, default to not logging system prompts.

### 🚨 Blocking Issues

**1. Module-level `os.getenv()` bypasses the Settings pattern**

The PR reads `NADIRCLAW_CORS_ORIGINS`, `NADIRCLAW_LOG_SYSTEM_PROMPTS`, `NADIRCLAW_HSTS` at module/import level. The project's `Settings` class (`settings.py`) uses `@property` with dynamic `os.getenv()` calls specifically because the CLI `serve` command sets `os.environ[...]` *after* import (`cli.py:57-69`). Module-level reads capture the empty default and silently ignore CLI overrides.

**Fix:** Add these as `@property` entries on the `Settings` class in `settings.py`, and read them via `settings.cors_origins` etc. at call time — not at import time.

**2. ThreadPoolExecutor logging — silent exception loss + shutdown data loss**

`_log_executor.submit(_do_log)` is fire-and-forget:
- If SQLite write fails (disk full, corruption), the exception is silently swallowed — no one reads the future's result.
- On SIGTERM, pending log entries in the executor queue are dropped.

**Fix:** (a) Add a `done_callback` on the returned future to log exceptions. (b) Add `_log_executor.shutdown(wait=True)` in a FastAPI shutdown event handler.

### ⚠️ Non-blocking Issues

**3. `re.compile` inside `_extract_request_metadata` — recompiled every request**

`_API_KEY_PATTERN` is defined inside the function body, so it gets recompiled on every call. Move it to module level — it's a constant.

**4. `_log_request` console print ordering**

After the refactor, the console `logger.info(...)` at the bottom of `_log_request` runs synchronously on the main thread *before* the thread pool I/O completes. The console says "logged" before the write actually succeeds. Minor, but worth noting.

**5. `return JSONResponse(content=response_body)` — behavior change**

Previously `chat_completions` returned a plain dict (FastAPI auto-serializes). Now it returns `JSONResponse`. Not breaking for HTTP clients, but any test that calls the endpoint function directly and checks `response["choices"]` will get a `JSONResponse` object instead of a dict.

**6. Constant-time auth — marginal improvement**

`hmac.compare_digest` with variable-length tokens returns early in CPython (documented). The timing-safety is weaker than the comments suggest. Not harmful, but the comment "constant-time" is slightly misleading for the variable-length case.

### Summary

The CORS fix and security headers are solid. **Issues 1 and 2 are blocking** — module-level env reads are a real functional bug (settings silently ignored via CLI), and fire-and-forget logging can lose data on failure/shutdown. The rest are minor.

doramirdor:
@pradumna-gautam — confirmed, the initial review (2026-03-28) referenced files that don't exist on this branch. Apologies. The CORS regex fix in `14d7ac2` is solid, nicely done.

**However**, the second review on 2026-04-07 is based on the actual files in this PR — those issues are still outstanding:

### Still blocking

**1. Module-level `os.getenv()` in `server.py`**

```python
_cors_origins_raw = os.getenv("NADIRCLAW_CORS_ORIGINS", "")
if _cors_origins_raw:
    _cors_origins = [...]
    _cors_credentials = True
else:
    _cors_origins = []
    _cors_origin_regex = r"^https?://..."
```

This runs at import time. The `serve` CLI (`cli.py:57-69`) sets `os.environ[...]` *after* import, so `nadirclaw serve --cors-origins https://myapp.com` would be silently ignored. Move these reads into `Settings` properties and read at CORSMiddleware registration time — or build the middleware lazily in a `startup` handler.

**2. `_log_executor.submit()` is fire-and-forget**

Two real problems:
- If SQLite write fails (disk full, locked db), the exception is swallowed — no one reads the future.
- On SIGTERM/uvicorn shutdown, queued log entries are lost.

**Fix:**
```python
fut = _log_executor.submit(_do_log)
fut.add_done_callback(lambda f: f.exception() and logger.error("log write failed", exc_info=f.exception()))

# and in a FastAPI shutdown handler:
@app.on_event("shutdown")
def _flush_logs():
    _log_executor.shutdown(wait=True)
```

### Non-blocking

**3. `_API_KEY_PATTERN` recompiled per request** — defined inside `_extract_request_metadata`. Move to module level.

**4. Truncation hardcoded to 500 chars in `request_logger.log_request`** — should be configurable or respect an existing setting rather than silently truncating prompts.

Once (1) and (2) are addressed, this is good to merge. The security headers + CORS regex + sanitized validation errors + bounds validation are all solid wins.

doramirdor:
@pradumna-gautam — friendly ping. This PR has been dormant since 2026-04-07. Two reviews on the latest commit (`14d7ac2`) flagged real blockers. The core/CORS/headers/bounds work is solid, but **two blockers remain outstanding:**

**1. Module-level `os.getenv()` bypasses the `Settings` pattern**

`NADIRCLAW_CORS_ORIGINS`, `NADIRCLAW_LOG_SYSTEM_PROMPTS`, `NADIRCLAW_HSTS` are read at import time. The CLI `serve` command sets env vars *after* import (`cli.py:57-69`), so overrides are silently ignored. Move to `@property` on `Settings` — same pattern used elsewhere in the codebase.

**2. `ThreadPoolExecutor` logging loses data silently**

`_log_executor.submit(_do_log)` is fire-and-forget:
- SQLite write failures are swallowed (no `done_callback` logging exceptions)
- Pending entries are dropped on SIGTERM (no `executor.shutdown(wait=True)` in FastAPI shutdown handler)

**Reminder — earlier critical issues, please confirm these are fixed in the rebase:**
- Auth crash: `matched_user_data = candidate_data,` (trailing comma → 1-tuple → `AttributeError` on every authenticated request)
- `trained_model.pkl` shipped in repo (joblib.load → pickle → arbitrary code execution risk) — should be gitignored, hash pinned in source not sidecar
- CORS wildcard patterns were invalid (`"http://localhost:*"`) — latest commit `14d7ac2` looks like it addressed this with `allow_origin_regex` ✓

This PR is security-critical and the rest of the open PR stack (#31) depends on it. If you have bandwidth to address (1) and (2) and confirm the auth/pickle fixes, happy to re-review. Otherwise let us know and we can either take it over or close it.

---

doramirdor:
@pradumna-gautam — final notice. This PR has been dormant since 2026-04-07 (~3 weeks). Two reviews flagged the same two outstanding blockers on `14d7ac2` and they're still unaddressed:

1. **Module-level `os.getenv()`** — `NADIRCLAW_CORS_ORIGINS`, `NADIRCLAW_LOG_SYSTEM_PROMPTS`, `NADIRCLAW_HSTS` read at import time. CLI `serve` overrides silently ignored. Fix: `@property` on `Settings`.
2. **Fire-and-forget `_log_executor.submit()`** — silent exception loss + drops queued entries on SIGTERM. Fix: `done_callback` for exception logging + `executor.shutdown(wait=True)` in a FastAPI shutdown handler.

The CORS regex fix and security headers in this PR are genuinely valuable and #31 depends on it landing. Plan:

- **By 2026-05-13** — if you can address (1) and (2), I'll re-review and merge.
- **After 2026-05-13** — I'll either close as stale or take it over (carving out the CORS/headers/bounds wins into a fresh PR with `Co-Authored-By` attribution to you), depending on bandwidth.

Let me know either way. Thanks for the original work — the design direction is right; we just need to land it.

---

doramirdor (approved):
Re-review of `f68d643` (now merged with main as `9250faa`).

### ✅ Both blockers resolved

**1. Module-level `os.getenv()` → `Settings` properties.** `CORS_ORIGINS`, `LOG_SYSTEM_PROMPTS`, `HSTS`, and `LOG_PROMPT_TRUNCATE` now live on `Settings` as `@property` (verified at [settings.py:271-292](https://github.com/NadirRouter/NadirClaw/blob/9250faa/nadirclaw/settings.py#L271-L292)). `server.py:146` reads `settings.CORS_ORIGINS` at middleware-registration time, `:186` reads `settings.HSTS` per-response, `:363` reads `settings.LOG_SYSTEM_PROMPTS` per-call. CLI `serve` env overrides now flow through.

Bad values fall back gracefully: `NADIRCLAW_LOG_PROMPT_TRUNCATE=garbage` → 500 with no exception.

**2. Fire-and-forget executor → bounded.** [`server.py:332`](https://github.com/NadirRouter/NadirClaw/blob/9250faa/nadirclaw/server.py#L332) attaches `_on_log_done` to log exceptions instead of swallowing them, and the new shutdown handler at [`server.py:477-480`](https://github.com/NadirRouter/NadirClaw/blob/9250faa/nadirclaw/server.py#L477-L480) calls `_log_executor.shutdown(wait=True)` so queued writes drain on SIGTERM. Verified with a smoke test.

### ✅ Bundled non-blockers

- **`_API_KEY_PATTERN`** moved to module level ([server.py:291](https://github.com/NadirRouter/NadirClaw/blob/9250faa/nadirclaw/server.py#L291)) — no per-request recompile.
- **Truncation** is configurable via `NADIRCLAW_LOG_PROMPT_TRUNCATE` ([request_logger.py:134](https://github.com/NadirRouter/NadirClaw/blob/9250faa/nadirclaw/request_logger.py#L134)).

### ✅ Local + CI verification

- Full suite (excluding `test_server.py` which needs network): **612 passed** in 94s on Python 3.14.
- CI green on 3.10 / 3.11 / 3.12 ([run 25884507280](https://github.com/NadirRouter/NadirClaw/actions/runs/25884507280)).

### Follow-up (non-blocking, will land separately)

`@app.on_event("shutdown")` is FastAPI-deprecated in favor of lifespan context managers — fine for now, worth modernizing in a small cleanup PR.

Merging.

---

PR #36, opened 2026-04-18
feat: model pool weighted load balancing

snowspig:
## Summary

- Add **model pool** weighted load balancing for distributing requests across multiple equivalent models
- Pool configuration via `NADIRCLAW_MODEL_POOLS` environment variable
- Weighted random selection with configurable weights per model
- Integrated into `apply_routing_modifiers()` — applies after all other routing decisions

## Configuration

```env
NADIRCLAW_MODEL_POOLS=turbo=gemini-2.5-flash,10+gpt-4.1-nano,5;reasoning=gpt-5.2,8+claude-opus-4-6-20250918,4
```

Format: `pool_name=model,weight+model,weight;pool_name=model,weight+...`

## How it works

1. When a model is routed and belongs to a pool, weighted random selection picks from that pool
2. Skips pool selection for `complex` and `reasoning` tiers (models explicitly chosen by detection)
3. Only applies for load-balancing equivalent models in `simple`/`mid` tiers

## Testing

- 12 unit tests covering: env parsing, weighted selection, reverse lookup, edge cases
- All tests pass

## Use case

When running multiple equivalent models (e.g., different API providers for the same model, or multiple budget models), pools distribute load proportionally to their weights. This avoids hammering a single provider and provides basic load balancing.

### DIFF
--- a/nadirclaw/routing.py
+++ b/nadirclaw/routing.py
@@ -6,6 +6,8 @@ model aliases, context-window filtering, and session persistence.
 
 import hashlib
 import logging
+import os
+import random
 import re
 import time
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


doramirdor (commented):
Thanks for this — weighted load balancing across equivalent models is a reasonable addition. Leaving as comment (not a formal request-changes) since nothing here is blocking, but a few things worth addressing before merge.

### ⚠️ Issues

**1. `MODEL_POOLS` built at import time — `serve` env overrides ignored**

```python
MODEL_POOLS: Dict[str, List[Tuple[str, int]]] = _parse_model_pools()
```

The CLI `serve` command sets env vars *after* `nadirclaw.routing` has already imported (`cli.py:57-69`). So `nadirclaw serve --set NADIRCLAW_MODEL_POOLS=...` silently uses the empty default. Either:
- Move `_parse_model_pools()` inside `apply_routing_modifiers` (reparse per request — cheap) or cache via `@lru_cache` that can be cleared, or
- Expose `NADIRCLAW_MODEL_POOLS` as a `@property` on `Settings` and build the map lazily on first routing call.

**2. `select_from_pool` returns `""` on unknown pool**

```python
if not pool:
    logger.warning("Unknown model pool: %s", pool_name)
    return ""
```

Returning an empty string will propagate into routing and fail downstream with a confusing error. Since this is only reached from `get_pool_for_model(...)` (which already guaranteed the model is in the pool), the branch is unreachable — drop it, or raise `KeyError`.

**3. Skipping `complex`/`reasoning` tiers is reasonable but should be documented on the env var**

Operators configuring `reasoning=gpt-5.2,8+claude-opus-4-6,4` will be surprised that it has no effect. Add a note in the PR description / README that pools only apply to `simple`/`mid` routed requests.

### Nitpicks
- `random.randint(1, total_weight)` — fine. If you ever want reproducibility for tests, consider `random.Random(seed).choices(...)`.
- Test uses `@mock.patch.dict(os.environ, ...)` but also calls the *already-imported* `_parse_model_pools()` directly — good, side-steps the import-time issue, but a production integration test should verify the env picks up via `serve`.

The tests are thorough. Once (1) is fixed this should merge cleanly.

doramirdor (changes requested):
Re-review of `3dbe859` ("lazy-init model pools + raise KeyError on unknown pool").

- **(1) Import-time `MODEL_POOLS`** — fixed via `_ensure_pools_loaded()` with double-checked locking. Verified that setting `NADIRCLAW_MODEL_POOLS` after `import nadirclaw.routing` now works as expected.
- **(2) `select_from_pool` returning `""`** — now raises `KeyError` with a helpful "Available: [...]" message. Good.
- Bonus: `reload_pools()` is a nice addition for runtime reconfiguration.

### 🚨 Blocking — the fix broke the test suite

`tests/test_model_pool.py` was not updated for the new internals. **11 of 12 tests fail** on this branch (`pytest tests/test_model_pool.py` → `11 failed, 1 passed`). Three independent breakages:

**A. `_parse_model_pools()` signature changed from `Dict` → `Tuple[Dict, Dict]`**

```python
# tests/test_model_pool.py:20
assert _parse_model_pools() == {}                   # now returns ({}, {})
# tests/test_model_pool.py:25-27
pools = _parse_model_pools()                        # tuple, not dict
assert pools["turbo"] == [("gemini-2.5-flash", 10)] # TypeError
```

Every test in `TestParseModelPools` (6 tests) fails. Fix: unpack the tuple.

```python
pools, _reverse = _parse_model_pools()
```

**B. Test setup writes to symbols that no longer exist**

```python
# tests/test_model_pool.py:84-85, 117-122
routing_mod.MODEL_POOLS = test_pools
routing_mod._MODEL_TO_POOL = reverse_map
```

`MODEL_POOLS` and `_MODEL_TO_POOL` were removed in this commit; the new caches are `_MODEL_POOLS_CACHE` / `_MODEL_TO_POOL_CACHE`. The assignments silently create unread attributes, so `select_from_pool` and `get_pool_for_model` fall through to the lazy loader (which sees an empty env) and either raise `KeyError` or return `None`. All 4 `TestSelectFromPool` and `TestGetPoolForModel::test_model_in_pool` fail. Fix: write to the new cache names, or better — drive the tests through `NADIRCLAW_MODEL_POOLS` + `reload_pools()`, which exercises the real code path.

**C. `test_unknown_pool_returns_empty` contradicts the new API**

```python
def test_unknown_pool_returns_empty(self):
    self._setup_pools()
    assert select_from_pool("nonexistent") == ""
```

This test name and assertion are now wrong by design — the function raises `KeyError` instead. Replace with `pytest.raises(KeyError)`.

### Nits
- `_ensure_pools_loaded` returns `Tuple[Dict, Dict]` typed, but `_MODEL_POOLS_CACHE` / `_MODEL_TO_POOL_CACHE` are `Optional[...]`. After the lock block they're guaranteed non-`None`, but mypy will complain — consider local non-`Optional` rebindings or `cast()`.
- The PR description already calls out the `complex`/`reasoning` skip behavior — good. Worth adding a one-liner near the `NADIRCLAW_MODEL_POOLS` env var in the README so operators don't configure a `reasoning=...` pool expecting it to be honored.

Once the test file is updated to match the new internals, this is good to merge.

---

doramirdor (approved):
Re-review of `2203f90` ("update tests for _parse_model_pools tuple return + new cache names").

### ✅ All prior blockers resolved
- **A. Tuple unpacking** — every `_parse_model_pools()` call now does `pools, _reverse = _parse_model_pools()`. `test_multiple_pools` also asserts on the reverse map (`reverse["gemini-2.5-flash"] == "turbo"`) — nice bonus coverage.
- **B. Cache-symbol names** — tests now write to `_MODEL_POOLS_CACHE` / `_MODEL_TO_POOL_CACHE`, matching the refactored internals.
- **C. Unknown-pool semantics** — renamed to `test_unknown_pool_raises_keyerror` and uses `pytest.raises(KeyError)`.

`pytest tests/test_model_pool.py` → **12 passed**. The earlier 11/12 failures are all resolved.

LGTM — approving.

---

PR #38, opened 2026-04-18
feat: complex coding detection and enhanced reasoning markers

snowspig:
## Summary

- **Complex coding detection** (`detect_complex_coding()`) — identifies multi-file editing tasks from recent tool usage:
  - Heavy editing (3+ Edit/Write calls)
  - Tool combo patterns (Read + Edit + Bash)
  - Deep conversation depth
  - Coding task keywords
- **Code review detection** (`detect_code_review()`) — identifies code review/audit tasks for routing to capable models
- **Enhanced reasoning markers** — added Chinese keywords (一步步, 深入分析, 逻辑推理, 优缺点, etc.) and additional English patterns (diagnose root cause, architectural decision, etc.)

## New API

```python
# routing.py — new functions
detect_complex_coding(messages, message_count) -> {"is_complex": bool, "confidence": float, "signals": list}
detect_code_review(prompt) -> {"is_review": bool, "confidence": float, "signals": list}
```

## Testing

- 19 new unit tests: Chinese reasoning markers, complex coding patterns, code review detection
- All 79 existing routing tests still pass

### DIFF
--- a/nadirclaw/routing.py
+++ b/nadirclaw/routing.py
@@ -401,7 +401,7 @@ _REVIEW_MARKERS = re.compile(
 )
 
 
-def detect_code_review(prompt: str) -> Dict[str, Any]:
+def detect_code_review(prompt: str, system_message: str = "") -> Dict[str, Any]:
     """Detect code review/verification tasks.
 
     Returns {"is_review": bool, "confidence": float, "signals": list}.
@@ -409,7 +409,8 @@ def detect_code_review(prompt: str) -> Dict[str, Any]:
     confidence = 0.0
     signals: List[str] = []
 
-    if _REVIEW_MARKERS.search(prompt):
+    text = f"{system_message}\n{prompt}" if system_message else prompt
+    if _REVIEW_MARKERS.search(text):
         confidence = 0.90
         signals.append("review_keywords")
 
--- a/nadirclaw/routing.py
+++ b/nadirclaw/routing.py
@@ -212,7 +212,8 @@ def _count_agentic_cycles(messages: List[Any]) -> int:
 # ---------------------------------------------------------------------------
 
 _REASONING_MARKERS = re.compile(
-    r"\b("
+    r"("
+    # English markers
     r"step[- ]by[- ]step"
     r"|think (?:through|carefully|deeply|about)"
     r"|chain[- ]of[- ]thought"
@@ -231,7 +232,24 @@ _REASONING_MARKERS = re.compile(
     r"|work through"
     r"|break (?:this|it) down"
     r"|logical(?:ly)? (?:deduce|infer|conclude)"
-    r")\b",
+    r"|analyze why (?:this|the|it)"
+    r"|diagnose the (?:root )?cause"
+    r"|weigh (?:the )?(?:pros|cons|options|alternatives)"
+    r"|architectural (?:decision|choice)"
+    r"|design (?:a )?(?:system|architecture)"
+    # Chinese markers
+    r"|一步步"
+    r"|逐步分析"
+    r"|深入思考"
+    r"|深入分析"
+    r"|推理分析"
+    r"|逻辑推理"
+    r"|优缺点"
+    r"|对比分析"
+    r"|权衡.*优劣"
+    r"|分析.*利弊"
+    r"|批判性分析"
+    r")",
     re.IGNORECASE,
 )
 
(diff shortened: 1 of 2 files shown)


doramirdor (commented):
Adding Chinese reasoning markers is a nice addition, and complex-coding detection is useful. Few issues before merge.

### 🚨 Blocking

**1. Removed `\b` word boundaries around reasoning regex — will cause false positives**

```diff
- r"\b("
+ r"("
  ...
- r")\b"
+ r")"
```

The `\b` was intentional. Without it:
- `"think through"` now matches inside `"overthink through pain"` or URLs / code identifiers.
- `"work through"` matches inside `"network throughput"`.
- `"deduce"` matches inside `"deduced"` — OK here, but the principle breaks.

Chinese characters don't have word boundaries (so `\b` doesn't help them), but the fix is to keep `\b` for English patterns and drop it only for Chinese. Split into two regexes, or use `(?:\b|(?<=[\u4e00-\u9fff]))` at the start.

### ⚠️ Bugs

**2. Confidence can exceed 1.0 internally**

`detect_complex_coding` sums: `0.50 + 0.30 + 0.20 + 0.40 = 1.40`. You clamp with `min(confidence, 1.0)` on return — good. But the intermediate comparison `is_complex = confidence >= 0.50` uses the unclamped value, which is still fine. Not a bug, just worth a comment explaining the design (linear confidence accumulation, not probabilities).

**3. `detect_code_review` only scans `prompt` — misses system prompt**

In agentic workflows, "code review this PR" is often in the system message, not the user message. Follow the `detect_reasoning(prompt, system_message)` signature:

```python
def detect_code_review(prompt: str, system_message: str = "") -> Dict[str, Any]:
    text = f"{system_message}\n{prompt}"
    ...
```

### Nits
- `_CODING_KEYWORDS` is a list of regex-like patterns but most don't use regex features. `implement` vs `add.*feature` are different idioms — just use plain `re.search` on each pattern uniformly (which you do). Fine but mixing literal keywords with regex patterns in the same list is confusing.
- `_REVIEW_MARKERS` includes `"lint\s*check"` — quite narrow. `"run lint"` or `"check for style issues"` wouldn't match.
- Tests reference Chinese: good coverage. Nice to see `detect_reasoning("请一步步分析，权衡优劣，给出优缺点")` hitting marker_count >= 3.

Issue (1) is the main blocker.

doramirdor (commented):
Re-review of `d315e39` ("restore \b word boundaries for English reasoning markers").

- **(1) Blocking — `\b` word boundaries** — fixed cleanly by splitting into `_REASONING_MARKERS_EN` (with `\b`, case-insensitive) and `_REASONING_MARKERS_ZH` (no `\b`, since CJK has no word boundaries). Verified the false positives are gone:
  - `"I overthink through pain"` → 0 markers ✓
  - `"network throughput is high"` → 0 markers ✓
  - `"please think through this"` → 1 marker ✓ (still works)
  - `"work through this problem"` → 1 marker ✓ (still works)
- All 19 tests in `tests/test_complex_coding.py` still pass.
- Bonus: expanded Chinese markers (证明, 推导, 论证, 演绎, 归纳, 设计方案) — useful additions.

### ⚠️ Still outstanding from prior review

**(3) `detect_code_review` still ignores the system message**

```python
# nadirclaw/routing.py:404
def detect_code_review(prompt: str) -> Dict[str, Any]:
```

This wasn't addressed in the fix commit. In agentic flows the "review this PR" instruction lives in the system prompt — `detect_reasoning(prompt, system_message="")` already handles this. The asymmetric signature will lead to misses in real workflows. Suggest:

```python
def detect_code_review(prompt: str, system_message: str = "") -> Dict[str, Any]:
    text = f"{system_message}\n{prompt}"
    ...
```

…and add a regression test that puts the marker in `system_message`.

### Nits (carried over, non-blocking)
- `_REVIEW_MARKERS` includes `"lint\s*check"` but won't catch `"run lint"` or `"check for style issues"` — narrow phrasing.
- Mixed regex/literal patterns in `_CODING_KEYWORDS` is still a touch confusing but fine.

Otherwise looks good — the regex split is the correct fix, and the additional Chinese markers are well-chosen.

---

doramirdor (approved):
Re-review of `690dcbc` ("detect_code_review accepts system_message parameter").

### ✅ Prior outstanding item resolved
```python
def detect_code_review(prompt: str, system_message: str = "") -> Dict[str, Any]:
    ...
    text = f"{system_message}\n{prompt}" if system_message else prompt
    if _REVIEW_MARKERS.search(text):
```

Signature now matches `detect_reasoning(prompt, system_message="")`. Two regression tests added:
- `test_review_keyword_in_system_message` — marker in both
- `test_review_keyword_only_in_system` — marker in system only

`pytest tests/test_complex_coding.py` → **21 passed** (was 19; +2 new).

### Security re-check
- Regex `_REVIEW_MARKERS` unchanged — still safe
- `\b` word-boundary split for `_REASONING_MARKERS_EN` preserved
- `f"{system_message}\n{prompt}"` is a simple concat, no injection concern
- No new env vars, no new attack surface

LGTM. Approving.

---

Issue #32, opened 2026-04-01
feat: Per-model format adapters for improved Claude Code tool-call translation


curtitoo:
### Problem

NadirClaw's routing intelligence (DistilBERT classifier, agentic detection, reasoning detection) is excellent. But when routing Claude Code requests to non-Anthropic models, the format translation relies entirely on LiteLLM's generic conversion. This works for simple text completions but breaks on:

1. **Tool call ID format mismatch** — Anthropic uses `toolu_01...`, OpenAI requires `call_...`. LiteLLM doesn't always translate these, causing tool calls to be rejected in conversation history replay.

2. **Thinking token handling** — Gemini requires `thought_signatures` to be persisted and echoed back across turns. DeepSeek returns `reasoning_content` in a non-standard field. Currently these are silently dropped or mangled.

3. **Model-specific tool call formats** — Grok returns tool calls as XML (`<xai:function_call>`), not JSON. Weak/local models return malformed JSON that needs recovery.

4. **Tool name limits** — Some models truncate tool names, breaking Claude Code's tool resolution on response.

These issues are most visible in heavy agentic usage (Claude Code with hooks, skills, MCP servers, sub-agent teams) where tool-calling fidelity is critical.

### Proposal

Add an optional **per-model adapter layer** between the classifier and LiteLLM dispatch. Each adapter would handle model-specific format quirks:

```
Request → Classifier → Adapter.convertRequest() → LiteLLM → Provider
Response ← Adapter.convertResponse() ← LiteLLM ← Provider
```

**Proposed adapters:**

| Adapter | Handles |
|---|---|
| `GeminiAdapter` | Thought signature persistence, `functionCall` format, `thinkingBudget` parameter |
| `OpenAIAdapter` | `max_tokens` → `max_completion_tokens`, `reasoning_effort` injection, tool call ID prefix |
| `GrokAdapter` | XML function call parsing (`<xai:function_call>`) |
| `DeepSeekAdapter` | `reasoning_content` → thinking block translation |
| `LocalAdapter` | Malformed JSON tool call recovery, tool name truncation handling |

Adapters would be optional — if no adapter matches the target model, LiteLLM handles it as today (zero breaking change).

### Architecture sketch

```python
# nadirclaw/adapters/base.py
class BaseAdapter:
    def should_handle(self, model: str) -> bool: ...
    def convert_request(self, request: dict) -> dict: ...
    def convert_response(self, response) -> response: ...

[...]

Issue #3, opened 2026-02-21
Ollama Setup Flow Improvements


LiemFrans:
This thread tracks the recent setup wizard changes for Ollama host input and model discovery.

## Summary

The setup wizard now:
- Prompts `OLLAMA_API_BASE` immediately after selecting providers (when Ollama is selected).
- Accepts host:port input (e.g. `10.4.136.145:11434`) and normalizes it to `http://10.4.136.145:11434`.
- Uses that base URL for:
  - connectivity checks in **Credentials**
  - model fetch in **Fetching Available Models**
  - persisted `.env` value (`OLLAMA_API_BASE=...`)

---

## Issue 1 — Prompt order for Ollama host in setup wizard

**Title**: Prompt `OLLAMA_API_BASE` before Credentials/Fetch steps for Ollama

**Problem**
- Wizard previously asked for Ollama host too late (around model selection), which made earlier connectivity and fetch steps use defaults.

**Expected behavior**
- If provider selection includes `ollama`, prompt host immediately after provider selection.
- Continue into Credentials and Fetch steps with this chosen host.

**Acceptance criteria**
- Selecting only Ollama prompts `OLLAMA_API_BASE` before `Credentials` block.
- `Credentials` output references selected host.
- `Fetching Available Models` uses selected host.

**Status**: ✅ Done (in my branch)

---

## Issue 2 — Hostname input without scheme should work

**Title**: Normalize Ollama host input to valid URL

**Problem**
- Input like `10.4.136.145:11434` failed because no `http://` or `https://` prefix was provided.

**Expected behavior**
- Blank input defaults to `http://localhost:11434`.
- Input without scheme gets `http://` prepended.
- Trailing slash is removed.

**Acceptance criteria**
- `10.4.136.145:11434` becomes `http://10.4.136.145:11434`.
- `http://host:11434/` becomes `http://host:11434`.

**Status**: ✅ Done (in my branch)

---

## Issue 3 — Fetch Ollama models from selected host

**Title**: Use selected `OLLAMA_API_BASE` for `/api/tags` fetch

**Problem**
- Model fetch could still rely on localhost if host selection happened later or wasn’t threaded through all calls.

**Expected behavior**
- Ollama fetch endpoint must be `${OLLAMA_API_BASE}/api/tags`.
- Fetched models should populate free-tier choices when available.

**Acceptance criteria**
- Setup output shows fetch from selected host.
- “using defaults” only appears when selected host fetch fails.

[...]

PR #39, opened 2026-04-18
feat: Anthropic Messages API compatibility layer

snowspig:
## Summary

- Add **`/v1/messages` endpoint** — full Anthropic Messages API compatibility
- Tools using the Anthropic SDK (Claude Code, Cursor, etc.) can use NadirClaw as a transparent proxy
- All routing benefits (agentic detection, reasoning, context window, session caching) apply to Anthropic-format requests

## Architecture

New module `nadirclaw/anthropic_api.py` (separate from server.py for clean separation):

- `anthropic_to_openai_messages()` — Anthropic → OpenAI message conversion
- `openai_response_to_anthropic()` — OpenAI → Anthropic response conversion
- `anthropic_tools_to_openai()` — Tool definition conversion
- `build_anthropic_sse_events()` — SSE streaming in Anthropic format
- `POST /v1/messages` — Main endpoint, delegates to internal routing

## Format Conversions

| Anthropic | OpenAI |
|-----------|--------|
| `system` field | `role: "system"` message |
| `content` blocks (text, tool_use, tool_result) | Structured messages |
| `tool_use` blocks | `tool_calls` array |
| `tool_result` blocks | `role: "tool"` messages |
| SSE events (message_start, content_block_delta, etc.) | Standard SSE |

## Testing

- 20 unit tests covering: message conversion, tool conversion, response conversion, SSE events, text extraction

### DIFF
--- a/nadirclaw/anthropic_api.py
+++ b/nadirclaw/anthropic_api.py
@@ -33,6 +33,10 @@ logger = logging.getLogger("nadirclaw.anthropic_api")
 
 router = APIRouter()
 
+# Same limits as /v1/chat/completions (server.py)
+_MAX_CONTENT_LENGTH = 1_000_000  # 1 MB
+_MAX_ANTHROPIC_TOKENS = 128_000  # Cap for max_tokens parameter
+
 # Provider env-var mapping for Anthropic-compatible endpoints.
 # Format: provider_name → (api_base_env_var, api_key_env_var)
 _ANTHROPIC_COMPAT_PROVIDERS: Dict[str, Tuple[str, str]] = {
@@ -241,8 +245,14 @@ def anthropic_to_openai_messages(
 
 
 def anthropic_tools_to_openai(tools: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
-    """Convert Anthropic tool definitions to OpenAI function format."""
+    """Convert Anthropic tool definitions to OpenAI function format.
+
+    Note: Anthropic built-in tools (bash_20250124, text_editor_20250124, etc.)
+    have no ``input_schema`` and are silently dropped. Only tools with
+    ``type="custom"`` or an explicit ``input_schema`` are forwarded.
+    """
     result = []
+    dropped = []
     for tool in tools:
         if tool.get("type") == "custom" or "input_schema" in tool:
             result.append({
@@ -253,6 +263,13 @@ def anthropic_tools_to_openai(tools: List[Dict[str, Any]]) -> List[Dict[str, Any
                     "parameters": tool.get("input_schema", {}),
                 },
             })
+        else:
+            dropped.append(tool.get("name", tool.get("type", "?")))
+    if dropped:
+        logger.warning(
+            "Dropped %d Anthropic built-in tools (no input_schema): %s",
+            len(dropped), dropped,
+        )
     return result
 
 
@@ -715,6 +732,17 @@ async def anthropic_messages(raw_request: Request):
     if settings.AUTH_TOKEN and effective_token != settings.AUTH_TOKEN:
         raise HTTPException(status_code=401, detail="Invalid API key")
 
+    # --- Rate limiting (parity with /v1/chat/completions) ---
+    from nadirclaw.server import _rate_limiter
+    user_id = effective_token or "anonymous"
+    retry_after = _rate_limiter.check(user_id)
+    if retry_after is not None:
+        raise HTTPException(
+            status_code=429,
+            detail=f"Rate limit exceeded. Retry after {retry_after}s.",
+            headers={"Retry-After": str(retry_after)},
+        )
+
     # Extract Anthropic fields
     ant_model = body.get("model", "")
     # ALL claude-* models → "auto" for smart routing (proven approach from 0.11.0)
@@ -724,6 +752,14 @@ async def anthropic_messages(raw_request: Request):
     ant_max_tokens = body.get("max_tokens", 4096)
     ant_tools = body.get("tools", [])
 
+    # Cap max_tokens to prevent abuse
+    if ant_max_tokens > _MAX_ANTHROPIC_TOKENS:
+        logger.warning(
+            "max_tokens=%d exceeds cap, clamping to %d",
+            ant_max_tokens, _MAX_ANTHROPIC_TOKENS,
+        )
+        ant_max_tokens = _MAX_ANTHROPIC_TOKENS
+
     prompt_text = _extract_last_user_text(body.get("messages", []))
 
     # Convert to OpenAI format
@@ -732,6 +768,19 @@ async def anthropic_messages(raw_request: Request):
         body.get("system"),
     )
     openai_tools = anthropic_tools_to_openai(ant_tools) if ant_tools else []
+
+    # --- Input size validation (parity with /v1/chat/completions) ---
+    total_content_len = sum(
+        len(m.get("content", "")) if isinstance(m.get("content"), str)
+        else len(json.dumps(m.get("content", "")))
+        for m in openai_messages
+    )
+    if total_content_len > _MAX_CONTENT_LENGTH:
+        raise HTTPException(
+            status_code=413,
+            detail=f"Request content too large ({total_content_len:,} chars). "
+                   f"Maximum is {_MAX_CONTENT_LENGTH:,} chars.",
+        )
 
     # Build ChatCompletionRequest
     chat_messages = []
--- a/nadirclaw/anthropic_api.py
+++ b/nadirclaw/anthropic_api.py
@@ -943,7 +943,22 @@ async def anthropic_messages(raw_request: Request):
             final_model = selected_model
 
             # Build candidate list: primary + fallback chain
+            # When tools are present, prefer direct-Anthropic candidates (better
+            # tool-call schema adherence) over LiteLLM candidates like gpt-5.4
+            # which may mangle parameter names (e.g., taskId vs task_id).
+            has_tools = bool(openai_tools)
             candidates = [selected_model] + fallback_chain
+            if has_tools:
+                anthropic_candidates = []
+                litellm_candidates = []
+                for c in candidates:
+                    cp = detect_provider(c)
+                    ce = get_anthropic_compat_endpoint(cp) if cp else None
+                    if ce:
+                        anthropic_candidates.append(c)
+                    else:
+                        litellm_candidates.append(c)
+                candidates = anthropic_candidates + litellm_candidates
 
             for candidate_model in candidates:
                 candidate_provider = detect_provider(candidate_model)
--- a/nadirclaw/anthropic_api.py
+++ b/nadirclaw/anthropic_api.py
@@ -19,6 +19,7 @@ the upstream model, then emits Anthropic SSE events.
 import json
 import logging
 import os
+import re
 import time
 import uuid
 from typing import Any, Dict, List, Optional, Tuple, Union
@@ -159,6 +160,21 @@ async def call_anthropic_direct(
         )
 
     return resp.json()
+
+
+# Pattern to strip system-reminder tags from model responses
+_SYSTEM_REMINDER_RE = re.compile(
+    r"<system-reminder>.*?</system-reminder>",
+    re.DOTALL,
+)
+
+
+def _sanitize_response_text(response: Dict[str, Any]) -> Dict[str, Any]:
+    """Strip <system-reminder> blocks from text content in a response."""
+    for block in response.get("content", []):
+        if block.get("type") == "text" and block.get("text"):
+            block["text"] = _SYSTEM_REMINDER_RE.sub("", block["text"]).strip()
+    return response
 
 
 def anthropic_response_to_stats(data: Dict[str, Any]) -> Dict[str, Any]:
@@ -940,6 +956,7 @@ async def anthropic_messages(raw_request: Request):
             # Try direct Anthropic call first, then fallback chain
             raw_response = None
             fallback_from = None
+            fallback_reasons: list[dict[str, str]] = []
             final_model = selected_model
 
             # Build candidate list: primary + fallback chain
@@ -982,10 +999,16 @@ async def anthropic_messages(raw_request: Request):
                             fallback_from = selected_model
                         break
                     except Exception as e:
+                        err_msg = str(e)[:200]
                         logger.warning(
                             "Direct Anthropic call failed for %s: %s — trying next",
-                            candidate_model, str(e)[:200],
+                            candidate_model, err_msg,
                         )
+                        fallback_reasons.append({
+                            "model": candidate_model,
+                            "reason": err_msg,
+                            "path": "direct_anthropic",
+                        })
                         continue
                 else:
                     # Path B: Convert to OpenAI and use LiteLLM (single try).
@@ -1022,6 +1045,7 @@ async def anthropic_messages(raw_request: Request):
                         "selected_model": final_model,
                         "tier": tier,
                         "fallback_used": fallback_from,
+                        "fallback_reasons": fallback_reasons or None,
                         "total_latency_ms": elapsed_ms,
                         **stats,
                         "status": "ok",
@@ -1071,6 +1095,7 @@ async def anthropic_messages(raw_request: Request):
                 "selected_model": final_model,
                 "tier": tier,
                 "fallback_used": fallback_from,
+                "fallback_reasons": fallback_reasons or None,
                 "total_latency_ms": elapsed_ms,
                 "prompt_tokens": prompt_tokens,
                 "completion_tokens": stats["completion_tokens"],
--- a/nadirclaw/anthropic_api.py
+++ b/nadirclaw/anthropic_api.py
@@ -1,0 +1,437 @@
+"""Anthropic Messages API compatibility layer for NadirClaw.
+
+Provides /v1/messages endpoint that accepts Anthropic SDK format requests,
+converts them to internal OpenAI format, routes through NadirClaw's smart
+routing, and returns responses in Anthropic format.
+
+This enables tools that use the Anthropic SDK (e.g., Claude Code) to work
+with NadirClaw as a transparent proxy.
+"""
+
+import json
+import logging
+import time
+import uuid
+from typing import Any, Dict, List, Optional, Union
+
+from fastapi import APIRouter, HTTPException, Request
+from fastapi.responses import JSONResponse
+
+from nadirclaw.settings import settings
+
+logger = logging.getLogger("nadirclaw.anthropic_api")
+
+router = APIRouter()
+
+
+# ---------------------------------------------------------------------------
+# Format conversion: Anthropic → OpenAI
+# ---------------------------------------------------------------------------
+
+def anthropic_to_openai_messages(
+    messages: List[Dict[str, Any]],
+    system: Optional[Union[str, List[Dict[str, Any]]]] = None,
+) -> List[Dict[str, Any]]:
+    """Convert Anthropic Messages API format to OpenAI chat completion format."""
+    result: List[Dict[str, Any]] = []
+
+    # System prompt
+    if system:
+        if isinstance(system, str):
+            result.append({"role": "system", "content": system})
+        elif isinstance(system, list):
+            text_parts = [
+                b.get("text", "")
+                for b in system
+                if isinstance(b, dict) and b.get("type") == "text"
+            ]
+            if text_parts:
+                result.append({"role": "system", "content": "\n".join(text_parts)})
+
+    for msg in messages:
+        role = msg.get("role", "user")
+        content = msg.get("content", "")
+
+        if role == "assistant":
+            if isinstance(content, list):
+                text_parts = []
+                tool_calls = []
+                for i, block in enumerate(content):
+                    if not isinstance(block, dict):
+                        continue
+                    if block.get("type") == "text":
+                        text_parts.append(block.get("text", ""))
+                    elif block.get("type") == "tool_use":
+                        tool_calls.append({
+                            "id": block.get("id", f"call_{i}"),
+                            "type": "function",
+                            "function": {
+                                "name": block.get("name", ""),
+                                "arguments": json.dumps(block.get("input", {})),
+                            },
+                        })
+                entry: Dict[str, Any] = {
+                    "role": "assistant",
+                    "content": "\n".join(text_parts) if text_parts else None,
+                }
+                if tool_calls:
+                    entry["tool_calls"] = tool_calls
+                result.append(entry)
+            else:
+                result.append({"role": "assistant", "content": content})
+
+        elif role == "user":
+            if isinstance(content, list):
+                text_parts = []
+                for block in content:
+                    if not isinstance(block, dict):
+                        continue
+                    if block.get("type") == "text":
+                        text_parts.append(block.get("text", ""))
+                    elif block.get("type") == "tool_result":
+                        tool_content = block.get("content", "")
+                        if isinstance(tool_content, list):
+                            tc_texts = [
+                                tc.get("text", "")
+                                for tc in tool_content
+                                if isinstance(tc, dict) and tc.get("type") == "text"
+                            ]
+                            tool_content = "\n".join(tc_texts)
+                        result.append({
+                            "role": "tool",
+                            "content": str(tool_content),
+                            "tool_call_id": block.get("tool_use_id", ""),
+                        })
+                if text_parts:
+                    result.append({"role": "user", "content": "\n".join(text_parts)})
+            else:
+                result.append({"role": "user", "content": content})
+        else:
+            result.append({"role": role, "content": str(content) if content else ""})
+
+    return result
+
+
+def anthropic_tools_to_openai(tools: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """Convert Anthropic tool definitions to OpenAI function format."""
+    result = []
+    for tool in tools:
+        if tool.get("type") == "custom" or "input_schema" in tool:
+            result.append({
+                "type": "function",
+                "function": {
+                    "name": tool.get("name", ""),
+                    "description": tool.get("description", ""),
+                    "parameters": tool.get("input_schema", {}),
+                },
+            })
+    return result
+
+
+# ---------------------------------------------------------------------------
+# Format conversion: OpenAI → Anthropic
+# ---------------------------------------------------------------------------
+
+def openai_response_to_anthropic(
+    response_data: Dict[str, Any],
+    model: str,
+    request_id: str,
+) -> Dict[str, Any]:
+    """Convert internal OpenAI-style response to Anthropic Messages API format."""
+    content_blocks = []
+
+    text = response_data.get("content")
+    if text:
+        content_blocks.append({"type": "text", "text": text})
+
+    for tc in response_data.get("tool_calls", []):
+        func = tc.get("function", {})
+        try:
+            input_data = json.loads(func.get("arguments", "{}"))
+        except (json.JSONDecodeError, TypeError):
+            input_data = {}
+        content_blocks.append({
+            "type": "tool_use",
+            "id": tc.get("id", str(uuid.uuid4())),
+            "name": func.get("name", ""),
+            "input": input_data,
+        })
+
+    if not content_blocks:
+        content_blocks.append({"type": "text", "text": ""})
+
+    finish = response_data.get("finish_reason", "stop")
+    if finish == "tool_calls":
+        stop_reason = "tool_use"
+    elif finish == "length":
+        stop_reason = "max_tokens"
+    else:
+        stop_reason = "end_turn"
+
+    return {
+        "id": f"msg_{request_id}",
+        "type": "message",
+        "role": "assistant",
+        "model": model,
+        "content": content_blocks,
+        "stop_reason": stop_reason,
+        "stop_sequence": None,
+        "usage": {
+            "input_tokens": response_data.get("prompt_tokens", 0),
+            "output_tokens": response_data.get("completion_tokens", 0),
+        },
+    }
+
+
+def build_anthropic_sse_events(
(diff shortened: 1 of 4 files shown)


doramirdor (commented):
Anthropic Messages API compat is genuinely valuable — Claude Code and Cursor would both benefit. But this implementation has auth and integration issues that are blocking.

### 🚨 Blocking

**1. Custom auth bypasses `validate_local_auth` — breaks multi-user + `X-API-Key` header**

```python
auth_header = raw_request.headers.get("authorization", "")
token = auth_header.replace("Bearer ", "") if auth_header.startswith("Bearer ") else ""
if settings.AUTH_TOKEN and token != settings.AUTH_TOKEN:
    raise HTTPException(status_code=401, detail="Invalid API key")
```

Three real problems:

- **Anthropic SDK sends `x-api-key: sk-ant-...`, not `Authorization: Bearer`.** The official `anthropic` Python SDK never sets an Authorization header. This endpoint will 401 every real Claude Code request.
- `validate_local_auth` already handles both headers and the `_LOCAL_USERS` multi-user dict (`auth.py:67-105`). Bypassing it means users configured via `credentials.json` or the users registry can't hit `/v1/messages`.
- Direct `!=` comparison is timing-attack-susceptible. `hmac.compare_digest` (already imported by the PR #30 branch) is the fix.

**Fix:** Use `Depends(validate_local_auth)` like `/v1/chat/completions`. It handles all three headers (Authorization, X-API-Key, x-api-key case-insensitive via FastAPI).

**2. `response_data` shape assumption is inconsistent with `_call_with_fallback`**

```python
response_data, selected_model, analysis_info = await _call_with_fallback(...)
...
text = response_data.get("content")
```

`_call_with_fallback` in `server.py` returns data matching OpenAI's chat completion shape — the response content lives under `choices[0].message.content`, not a top-level `content` key. Either this is silently producing empty responses, or I'm missing something. Please add an integration test that actually hits `/v1/messages` with a real LLM call and verifies the response body is non-empty.

### ⚠️ Design

**3. Per-request imports from `server.py`**

```python
async def anthropic_messages(raw_request: Request):
    from nadirclaw.server import (
        _call_with_fallback, _extract_request_metadata, _log_request, ...
    )
    from nadirclaw.server import app  # never used
```

Python caches in `sys.modules` so it's not slow, but this reads as a circular-import workaround. Either resolve the circular dependency structurally (move shared helpers to a third module) or at least hoist to module level with a conditional import. The unused `app` import should be removed.

**4. `settings.AUTH_TOKEN` is still module-level-ish**

`Settings.AUTH_TOKEN` is a `@property` that reads env dynamically — good. But it's used at function-call time here, which is fine. This isn't a bug, just clarifying since I almost flagged it.

**5. No cost tracking / prompt caching passthrough**

`/v1/chat/completions` goes through `_extract_request_metadata` + routing + `_call_with_fallback` which handles pricing and cache headers. The Anthropic path does routing but the `_log_request` call omits `cost_usd`, `cached_tokens`, etc. Parity gap — users who migrate from `/v1/chat/completions` to `/v1/messages` lose billing visibility.

### Nits
- `anthropic_tools_to_openai` skips anything without `input_schema` — but Anthropic also supports `type: "bash_20250124"` and `type: "text_editor_20250124"` built-in tools. Those will silently drop. Document this limitation or map them explicitly.
- `tool_use_id` in Anthropic is mapped to `tool_call_id` in OpenAI — correct.
- SSE format looks correct (`message_start`, `content_block_delta`, `message_stop`). Nice work on the event ordering.
- `_extract_last_user_text` gracefully handles blocks with only `tool_result` — returning `""` is the right call.

This is a substantial piece of work and the conversion logic is well-tested in unit tests. The blockers are (1) the auth bug which I'm confident fails in production, and (2) the response-shape concern which needs an integration test to confirm. Once those are resolved this is a great addition.

doramirdor (commented):
Re-review of `b6bdd78` ("use validate_local_auth + fix response shape + cost tracking").

- **(1) Custom auth bypassed `validate_local_auth`** — fixed. Now uses `current_user: UserSession = Depends(validate_local_auth)`. This gives you `X-API-Key` support (alias match is case-insensitive in FastAPI, so `x-api-key` from the Anthropic SDK works), multi-user `_LOCAL_USERS` lookup, and the `compare_digest`-style equality check that `validate_local_auth` already does. Real Claude Code requests should now authenticate.
- **(2) Response-shape mismatch** — fixed. Unpacks `choices[0].message.{content,tool_calls}` and `usage.{prompt_tokens,completion_tokens}` before handing to `openai_response_to_anthropic`. Matches what `_call_with_fallback` actually returns.
- **(3) Unused `app` import** — removed. `Depends` is hoisted to the module-level FastAPI import.
- **(5) Cost/cache parity with `/v1/chat/completions`** — `_log_request` now includes `cost_usd`, `cached_tokens`, `message_count`, `has_system_prompt`, `system_prompt_length`, `tool_count`, `has_tools`, `requested_model`, `stream`. Good parity.

All 20 unit tests pass (`pytest tests/test_anthropic_api.py` → `20 passed`).

### ⚠️ Still outstanding

**A. No integration test for the actual `/v1/messages` endpoint**

The prior review asked for a TestClient-based test that POSTs to `/v1/messages` with a mocked `_call_with_fallback` and verifies the response body is non-empty and in Anthropic format. This would have caught the original `response_data.get("content")` bug. All 20 current tests exercise only the conversion helpers in isolation — none go through the actual handler, auth dependency, or log path. Please add at least one, e.g.:

```python
from fastapi.testclient import TestClient
from nadirclaw.server import app

def test_messages_endpoint_non_empty(mocker):
    mocker.patch("nadirclaw.server._call_with_fallback",
                 return_value=({"choices": [{"message": {"content": "hi"}, "finish_reason": "stop"}],
                                "usage": {"prompt_tokens": 4, "completion_tokens": 1}},
                               "gpt-4.1-mini", {"cost_usd": 0.001}))
    client = TestClient(app)
    r = client.post("/v1/messages",
                    json={"model": "claude-sonnet-4-6", "max_tokens": 16,
                          "messages": [{"role": "user", "content": "hi"}]})
    assert r.status_code == 200
    body = r.json()
    assert body["content"][0]["text"] == "hi"
    assert body["role"] == "assistant"
```

**B. Per-request imports from `nadirclaw.server` still live inside the handler**

```python
from nadirclaw.server import (
    _call_with_fallback, _extract_request_metadata, _log_request, _rate_limiter,
    ChatMessage, ChatCompletionRequest,
)
```

The circular dependency this works around is real (`server.py` mounts `router` from this module), but this shape is fragile — every call pays the `sys.modules` lookup and `importlib` lock. Extract the shared helpers into `nadirclaw/_internal.py` (or similar), have both `server.py` and `anthropic_api.py` import from there, and hoist the imports to module scope. Non-blocking but worth a follow-up.

**C. `anthropic_tools_to_openai` still silently drops built-in tools**

```python
if tool.get("type") == "custom" or "input_schema" in tool:
    result.append({...})
# else: dropped on the floor
```

Anthropic SDK users frequently declare `{"type": "bash_20250124", "name": "bash"}` and `{"type": "text_editor_20250124", "name": "str_replace_editor"}`. These have no `input_schema` and are silently dropped. Either log a warning, or map them to a reasonable function stub so the model doesn't hallucinate tool calls that never route. At minimum document the limitation in the function docstring.

### Nits (still applicable)
- `ant_max_tokens = body.get("max_tokens", 4096)` is never passed to `_call_with_fallback` — Anthropic clients that set a lower `max_tokens` (common in Claude Code for cost caps) will get the provider default instead.
- Stream branch (`if ant_stream:`) still uses `build_anthropic_sse_events` on the already-collapsed `response_data`, not on the streaming iterator — true token-by-token streaming isn't wired through. This wasn't in the prior review but worth flagging as a follow-up since the Anthropic SDK expects real SSE.

### Summary

The two blocking issues from the prior review are genuinely fixed and tests pass. What's left is (a) one integration test to catch shape regressions, (b) a structural cleanup for the circular import, and (c) built-in tool handling. None of these should block merge if you're willing to take them as follow-ups — I'd push back mainly on (a) since the prior bug would have been caught by it. Flagging as a comment rather than request-changes.

---

doramirdor (changes requested):
Security review before merge — **⚠️ holding off: two authenticated-bypass issues**.

### 🚨 Security: `/v1/messages` skips protections applied to `/v1/chat/completions`

**1. Rate limiting is not enforced**

```python
from nadirclaw.server import (
    _call_with_fallback,
    _extract_request_metadata,
    _log_request,
    _rate_limiter,               # ← imported
    ChatMessage,
    ChatCompletionRequest,
)
# ...but never called on this path
```

Compare to `server.py:1089`:
```python
retry_after = _rate_limiter.check(current_user.id)
if retry_after is not None:
    raise HTTPException(status_code=429, ...)
```

An authenticated user can bypass per-user rate limits entirely by calling `/v1/messages` instead of `/v1/chat/completions`. For a hosted deployment this means cost-amplification (any user can saturate the provider quota), and in single-tenant setups it still defeats the per-model RPM caps set with `NADIRCLAW_RATE_LIMIT`. Fix: add the same `_rate_limiter.check(current_user.id)` + 429 response before dispatch.

**2. Input size validation is missing**

`/v1/chat/completions` has:
```python
total_content_len = sum(len(m.text_content()) for m in request.messages)
if total_content_len > _MAX_CONTENT_LENGTH:
    raise HTTPException(status_code=413, ...)
```

`/v1/messages` reads `body.get("messages", [])`, `body.get("system")`, `body.get("tools", [])` with no caps. An authenticated user can POST multi-MB payloads that consume memory during conversion, run through the full routing pipeline, and then either get rejected upstream or silently tax the proxy. Also: `ant_max_tokens = body.get("max_tokens", 4096)` has no ceiling and isn't forwarded anyway — dead code that should either be enforced or removed.

Fix: run the same `_MAX_CONTENT_LENGTH` check after `openai_messages` is built, or before JSON parsing if you want to reject oversize bodies earlier.

### Why this matters

PR #30's whole point was adding rate-limit and bounds validation on the single hot endpoint. A new endpoint that accepts the same workloads needs to enforce the same protections, otherwise PR #30 is defeated for any client that can switch protocols. This is a straightforward defense-in-depth fix — ~6 lines.

### Other (non-security) items still open
- Integration test for `/v1/messages` (previously flagged)
- Circular import via function-scoped imports (previously flagged)
- Built-in Anthropic tools silently dropped (`bash_20250124`, `text_editor_20250124`)

### Decision

Not merging until (1) and (2) are resolved. The conversion code looks correct and tests pass — just need parity with `/v1/chat/completions` on the protection surface.

---

doramirdor (changes requested):
Re-review after `7da19e7` + `a9f61ee`.

### ✅ My previous security blockers are partially addressed
- **Rate limiting** — now calls `_rate_limiter.check(user_id)` → 429. Good.
- **Input size validation** — `_MAX_CONTENT_LENGTH = 1_000_000` enforced after Anthropic→OpenAI conversion → 413.
- **`max_tokens` cap** — clamped at 128K. Good.
- **Built-in tools** — dropped with `logger.warning` + name list. 24/24 tests pass.

### 🚨 New security regression in `7da19e7` — prior auth fix was reverted

My re-review of `b6bdd78` approved this handler signature:
```python
@router.post("/v1/messages")
async def anthropic_messages(
    raw_request: Request,
    current_user: UserSession = Depends(validate_local_auth),
):
```

`7da19e7` reverted to inline custom auth:
```python
auth_header = raw_request.headers.get("authorization", "")
token = auth_header.replace("Bearer ", "") if auth_header.startswith("Bearer ") else ""
x_api_key = raw_request.headers.get("x-api-key", "")
effective_token = token or x_api_key
if settings.AUTH_TOKEN and effective_token != settings.AUTH_TOKEN:
    raise HTTPException(status_code=401, detail="Invalid API key")
```

This reintroduces the original PR's security defects:

**1. Timing-attack-susceptible equality.** `effective_token != settings.AUTH_TOKEN` short-circuits character-by-character. `validate_local_auth` / `hmac.compare_digest` exists for this reason.

**2. `_LOCAL_USERS` multi-user registry is bypassed.** Users configured via `~/.nadirclaw/credentials.json` (or `auth.py:_LOCAL_USERS`) have per-user tokens. `/v1/chat/completions` accepts them; `/v1/messages` rejects everything except the single global `settings.AUTH_TOKEN`. In a multi-user deployment, the new endpoint is unreachable for legitimate users.

**3. Missing 1000-char token length cap.** `validate_local_auth` rejects oversized tokens at line 86 (`_MAX_TOKEN_LENGTH = 1000`). The inline version accepts arbitrarily long header values — memory abuse surface.

**4. Rate-limit bypass in no-auth mode.** With `settings.AUTH_TOKEN` unset (local-only mode), any token string passes auth. `user_id = effective_token or "anonymous"` then uses *the attacker-supplied token* as the rate-limit key — rotating through random tokens gives a fresh bucket each request, completely defeating `_rate_limiter.check`. Compare to `/v1/chat/completions` which uses `current_user.id` from `UserSession`, which is a stable identity regardless of token presence.

**Fix:** put back `current_user: UserSession = Depends(validate_local_auth)` and use `current_user.id` as the rate-limit key. If the direct-call path needs the raw bearer token to forward to the upstream provider (I don't think it does — the upstream uses `api_key` from env), extract that from `raw_request` but still go through `validate_local_auth` for the gatekeeper check.

### ⚠️ New surface to think about — direct Anthropic call path (~680 lines in `7da19e7`)

`call_anthropic_direct(api_base, api_key, model, body)` does an httpx POST to operator-configured provider endpoints. Quick security walk:
- `api_base` and `api_key` both come from env — operator trust boundary, fine.
- `url = f"{api_base}/v1/messages"` — no user influence.
- Request body fields all come from the JSON body. The `anthropic_beta` header is user-controlled:
  ```python
  beta = body.get("anthropic_beta", "")
  if beta:
      headers["anthropic-beta"] = beta
  ```
  httpx validates header values for CRLF so smuggling is blocked at the library layer, but worth adding an explicit sanity check (strip non-ASCII or bound length) since you're forwarding user input to a privileged header.
- `resp.text[:500]` bounded — good, no unbounded error echo.

Non-security concerns, but worth calling out given the size of this commit: the added path duplicates a lot of the routing/logging pipeline that `/v1/chat/completions` already has. It deserves its own integration test that actually exercises the whole `anthropic_messages` → `call_anthropic_direct` flow through a mocked httpx transport, given the original bugs were caught by manual inspection rather than tests.

### Decision

Can't approve yet — the auth regression is a genuine step backward. Flipping back to `Depends(validate_local_auth)` + switching the rate-limit key to `current_user.id` should be ~10 lines.

Once that's resolved, I'll re-review and approve if clean.

---

doramirdor (changes requested):
Re-review after `86633ea`, `00f6cec`, `9db18fe`, `047288f`, `20cb603` (5 new commits since last review).

### 🚨 The auth regression I flagged is still present

```python
# nadirclaw/anthropic_api.py:770-780 (unchanged from a9f61ee)
auth_header = raw_request.headers.get("authorization", "")
token = auth_header.replace("Bearer ", "") if auth_header.startswith("Bearer ") else ""
x_api_key = raw_request.headers.get("x-api-key", "")
effective_token = token or x_api_key
if settings.AUTH_TOKEN and effective_token != settings.AUTH_TOKEN:
    raise HTTPException(status_code=401, detail="Invalid API key")
...
user_id = effective_token or "anonymous"
retry_after = _rate_limiter.check(user_id)
```

All four issues from the previous review still apply: timing-unsafe `!=`, `_LOCAL_USERS` bypass, no token length cap, rate-limit bucket bypass via token rotation in no-auth mode. None of the 5 new commits touched the auth path.

### 🚨 32 existing routing tests now fail

`pytest tests/test_routing.py` → **32 failed, 47 passed**. Examples:

- `TestDetectReasoning::test_reasoning_in_system` — broken by `86633ea` ("reasoning detection ignores system message"). The change is backward-incompatible and the test in main asserts the prior behavior.
- `TestResolveAlias::test_sonnet` — `resolve_alias("sonnet")` returns `claude-sonnet-4-6` instead of `claude-sonnet-4-5-20250929` (alias map changed without test update).
- `TestEstimateCost::test_known_model` — `estimate_cost("gpt-4o", ...)` now returns `None` (registry entry removed/renamed).
- 29 more across `TestDetectAgentic`, `TestApplyRoutingModifiers`, `TestVisionModifier`, `TestThreeTierClassifier`, `TestSessionCache`, etc.

This PR is in a state where merging will red the build for everyone working off main.

### 🚨 Scope sprawl — this is no longer just an Anthropic API compat PR

```
$ git diff main --stat
nadirclaw/anthropic_api.py  | 1136 ++++++++++++++++++++++++++++++++++++++++
nadirclaw/routing.py        | 1149 +++++++++++++++++++++++++++++++++++++---
nadirclaw/server.py         | 1215 ++++++++++++++++++++++++++++++++++++++++++-
tests/test_anthropic_api.py |  252 +++++++++
4 files changed, 3667 insertions(+), 85 deletions(-)
```

Concerns:

1. **`routing.py` includes the model-pool code from PR #36.** I merged #36 to `main` (commit `0178644`) ~1 hour ago. This branch needs a rebase to drop the duplicate pool implementation from `9db18fe`.

2. **Multiple Anthropic helpers are duplicated in `server.py`.**

   ```
   server.py defines:           anthropic_api.py already defines:
   _anthropic_to_openai_messages    anthropic_to_openai_messages
   _anthropic_tools_to_openai       anthropic_tools_to_openai
   _openai_response_to_anthropic    openai_response_to_anthropic
   _build_anthropic_streaming_response  (same)
   _extract_last_user_text          (same)
   _call_anthropic_direct           call_anthropic_direct
   _sanitize_anthropic_messages     (was mentioned in commit msg but not present in module)
   anthropic_messages               anthropic_messages (endpoint!)
   ```

   Two endpoint implementations of `/v1/messages` is going to cause a route conflict at startup or one will silently shadow the other. Pick one home — `anthropic_api.py` is the correct one based on the PR's stated architecture.

3. **`_heuristic_route` is a whole new routing strategy added to server.py** — this is a separate feature, not Anthropic-API-related. Belongs in its own PR with its own review.

4. **Logging changes (`047288f`, `20cb603`) — `<system-reminder>` filtering, response content logging, fallback reasons.** All useful but unrelated to the PR title. Should be split.

### Recommendation

This PR has effectively become a feature train. Before any further review:

1. **Rebase against current main** (which now includes #36's pool code). This will likely cut ~1000 lines from the diff.
2. **Drop the duplicated Anthropic helpers** from `server.py`. Keep them in `anthropic_api.py` only.
3. **Fix or update the 32 broken routing tests** — either revert the backward-incompatible changes or update the test expectations *and explain why* in the commit message.
4. **Address the auth regression** (use `Depends(validate_local_auth)` and `current_user.id` for the rate-limit key — ~10 lines).
5. **Consider splitting** the unrelated changes (`_heuristic_route`, system-reminder log filtering, response-text logging, fallback-reasons logging, model-pool-tools-priority) into their own PRs.

Holding **REQUEST_CHANGES** until at least (3) and (4) are resolved. (1)–(2) are strongly recommended for reviewability but not strictly merge-blocking once the tests pass.

---

doramirdor:
Thanks for the comprehensive PR @snowspig. The `/v1/messages` surface + direct-Anthropic passthrough path is genuinely valuable for Anthropic-SDK clients, and the "fake streaming" approach matches what we already ship in v0.11+. A few things need attention before this can merge — some critical.

## 🚨 Blocking

**1. `MODEL_REGISTRY` has 5 duplicate `"gpt-5.4"` keys — silently erases 5 OpenAI models**

```python
# routing.py, inside MODEL_REGISTRY
"gpt-5.4": {"context_window": 1_047_576, ... "cost_per_m_input": 2.00, ...},  # was gpt-4.1
"gpt-5.4": {"context_window": 1_047_576, ... "cost_per_m_input": 0.40, ...},  # was gpt-4.1-mini
"gpt-5.4": {"context_window": 1_047_576, ... "cost_per_m_input": 0.10, ...},  # was gpt-4.1-nano
"gpt-5.4": {"context_window": 128_000,   ... "cost_per_m_input": 2.50, ...},  # was gpt-4o
"gpt-5.4": {"context_window": 128_000,   ... "cost_per_m_input": 0.15, ...},  # was gpt-4o-mini
```

Python silently dedupes on dict construction, so only the **last** entry survives (the gpt-4o-mini pricing at $0.15 input / 128k context). `gpt-4.1`, `gpt-4.1-mini`, `gpt-4.1-nano`, `gpt-4o`, `gpt-4o-mini` are all erased from the registry. Users who explicitly route to those models will hit the default fallback pricing and context window.

Looks like a bulk find-replace (`gpt-4.1 → gpt-5.4`, `gpt-4o → gpt-5.4`) that collapsed five distinct models into one key. Please restore the original keys (`gpt-4.1`, `gpt-4.1-mini`, `gpt-4.1-nano`, `gpt-4o`, `gpt-4o-mini`) with their pricing, and add `gpt-5.4` as a separate row (once) if it's a real model you need.

Same issue in `MODEL_ALIASES`:
```python
"gpt-5.4": "gpt-5.4",   # self-referential
"gpt5.4": "gpt-5.4",
"gpt-5.4": "gpt-5.4",   # duplicate + self-referential
```
Self-referential aliases are no-ops; the duplicate key collapses. Also `"claude-sonnet-4-6": "claude-sonnet-4-6"` is a no-op — drop it.

**2. Scope / merge conflict**

- 3,667 additions across 4 files, in `CONFLICTING` state against current `main`.
- `routing.py` gets +1,086/-63 and `server.py` gets +1,193/-22, but the PR summary only describes the `/v1/messages` endpoint. The routing.py diff contains substantial unrelated work: `select_from_pool`, `detect_complex_coding`, `detect_code_review`, `detect_claude_code_role`, `_CONTEXT_INJECTION_PATTERNS`, Chinese reasoning markers, agentic threshold bump from 0.35 → 0.80, session cache key change (first user → last user). Much of this overlaps with #36 (model pool) and #38 (complex coding + Chinese markers) which already landed, which is likely why it's DIRTY. Please rebase onto current `main` and drop the overlapping code.

**Ask:** split this into at least two PRs:
1. `/v1/messages` endpoint + format conversions (the actual feature) — `anthropic_api.py` + minimal `server.py` wiring.
2. Anything still needed in `routing.py` after rebase, justified against what #36/#38 already provide.

**3. Rate-limit & helper reach-across into `server.py` private symbols**

```python
# anthropic_api.py:anthropic_messages
from nadirclaw.server import (
    _call_with_fallback, _dispatch_model, _extract_request_metadata,
    _log_request, _smart_route_full, ChatMessage, ChatCompletionRequest, UserSession,
)
from nadirclaw.server import _rate_limiter
```

Two separate `from nadirclaw.server import` statements inside the handler, importing 9 underscore-prefixed private helpers. This creates a tight circular dependency (`server.py` registers `router` from `anthropic_api`, which imports from `server`) and makes refactoring server.py impossible without breaking the endpoint silently.

Please extract the shared pieces (`_smart_route_full`, `_dispatch_model`, `_call_with_fallback`, `_extract_request_metadata`, `_log_request`, `_rate_limiter`, request models) into a neutral module (e.g. `nadirclaw/_core.py` or `nadirclaw/routing_pipeline.py`) that both `server.py` and `anthropic_api.py` can import cleanly.

**4. `get_anthropic_compat_endpoint` reads `.env` at every call**

```python
def get_anthropic_compat_endpoint(provider: str) -> Optional[Tuple[str, str]]:
    from dotenv import dotenv_values
    from pathlib import Path
    _env_file = Path.home() / ".nadirclaw" / ".env"
    file_vals = dotenv_values(_env_file) if _env_file.exists() else {}
    ...
```

Hitting the filesystem on every `/v1/messages` request (and repeatedly inside the candidate loop — it's called per-candidate) is a latency regression at load. `Settings` already loads `~/.nadirclaw/.env` once at import time; add `ZAI_API_BASE` / `KIMI_API_BASE` / `MINIMAX_API_BASE` / `ZAI_API_KEY` / … as `@property` on `Settings` and read from there.

Also, silently skipping when `api_key in ("local", "dummy", "sk-placeholder", "")` — that's OK as behavior, but please log at `INFO` the first time a placeholder is observed so users who misconfigure can tell why passthrough isn't happening.

## ⚠️ Bugs

**5. Token count path uses `max()` of reported vs. estimated — but estimate is naive**

```python
est_chars = len(json.dumps(body.get("system", "")))
est_chars += len(json.dumps(body.get("messages", [])))
est_chars += len(json.dumps(body.get("tools", [])))
estimated_pt = est_chars // 4
prompt_tokens = max(reported_pt, estimated_pt)
```

`json.dumps` on strings adds escaping + quotes, so this over-counts on heavy string content and under-counts on Unicode / CJK (where chars-per-token is closer to 1). For GLM/Kimi, where the reason for the fallback is "upstream reports suspiciously low values", `max()` at least avoids under-counting, but this will hit the accounting/pricing path downstream and over-bill users on multi-byte prompts. Please either (a) use a proper tokenizer (tiktoken fallback) or (b) cap the inflation (e.g., `max(reported_pt, min(estimated_pt, reported_pt * 2))`).

**6. `anthropic_tools_to_openai` drops Anthropic built-in tools with only a `logger.warning`**

```python
if tool.get("type") == "custom" or "input_schema" in tool:
    result.append(...)
else:
    dropped.append(tool.get("name", tool.get("type", "?")))
```

If a Claude Code client sends `bash_20250124` or `text_editor_20250124`, this silently drops them and LiteLLM returns "no tools" — the agent loop will hang or repeatedly re-request tool use. This is the exact "bug on weak fallback" scenario #32 is asking us to fix. At minimum, return `400` when the only tools in the request are built-in tools on a non-Anthropic route, so the caller gets a clear error instead of a silent no-op.

**7. `record_llm_call` only fires on the LiteLLM path**

In the direct-Anthropic path, `record_llm_call(span, ...)` is called after success. In the LiteLLM path, it's called *inside the candidate loop* before the response is returned, which means the telemetry span is recorded even when we then `return` mid-loop. Fine in isolation, but the direct path's `record_llm_call` is after the `for candidate_model in candidates:` loop ends, so if the first candidate succeeds we're still inside the `with trace_span(...)` block — which is OK. But if all candidates fail and we raise `RuntimeError("All models failed...")`, `record_llm_call` is never called and the span closes with no LLM metrics attached. Consider recording a failed-call metric in the `except` branch.

**8. `ant_stream = body.get("stream", False)` but `stream=False` is hardcoded downstream**

```python
req_data: Dict[str, Any] = {
    ...
    "stream": False,  # Always non-streaming internally
}
```

OK as a design (fake streaming), but please state this explicitly in the endpoint docstring so users know they never get token-by-token latency — it's a visible product characteristic.

## Non-blocking

- **`detect_agentic` threshold raised 0.35 → 0.80 via commented-out dead code.** Please delete the old scoring branches rather than leaving them commented; dead code rots. If you want the pre-Claude-Code behavior recoverable, gate it behind a setting.
- **`_REASONING_MARKERS_ZH` is great** but matches `分析.*利弊` with `.*` unanchored — on a long prompt this is O(n²) worst case. Consider `.{0,40}` or bounded anchors.
- **`is_main_session` check uses a literal substring** `"You are Claude Code, Anthropic's official CLI"` in three places in `routing.py`. Extract to a named constant `_CLAUDE_CODE_MAIN_SESSION_MARKER` at module top.
- **`anthropic_api.py:_extract_response_text` truncates to 500 chars** — fine for logging, but the function name doesn't say "for logging". Rename to `_format_response_for_log` or similar.
- **Test coverage is light** — 252 lines of tests for ~2.2k lines of new non-trivial behavior (routing logic changes, role detection, direct-Anthropic path, fallback chain). At minimum, add tests for: registry key collisions (would have caught #1), placeholder-key skip path, tool drop warning, the `last_message_is_tool` reasoning skip, and the `is_main_session` agentic skip.

## What to do next

1. Rebase onto `main` (you'll drop the overlap with #36/#38 automatically).
2. Fix the duplicate dict keys (#1) — this is a strict regression.
3. Extract shared pipeline symbols out of `server.py` into a neutral module (#3).
4. Split the non-endpoint routing changes into a separate PR, or justify each one against current `main` after rebase.

Happy to re-review once these land. The core feature (Anthropic Messages API passthrough) is good — the blockers are the refactor + dict-key regressions.

---

doramirdor:
@snowspig — friendly ping. This PR has been dormant since 2026-04-19 (~11 days). The latest commit (`20cb603`) still has the four blockers from my last review:

1. **Auth regression in `7da19e7`** — inline custom auth replaced `Depends(validate_local_auth)`. Reintroduces timing-unsafe `!=`, `_LOCAL_USERS` registry bypass, missing token-length cap, and rate-limit bucket bypass via token rotation in no-auth mode. Fix: put back `Depends(validate_local_auth)` and use `current_user.id` for the rate-limit key (~10 lines).

2. **32 routing tests fail on this branch** — `pytest tests/test_routing.py` → 32 failed, 47 passed. The backward-incompatible changes in `86633ea` (reasoning detection ignores system) and the alias/registry changes don't update the test expectations. Either revert or update with a justification.

3. **Scope sprawl + duplicate code** — 3,667 additions across 4 files. `routing.py` now contains the model-pool code from #36 (merged to main). `server.py` has duplicates of every Anthropic helper that already lives in `anthropic_api.py`, plus a second `/v1/messages` registration that will collide at startup. Rebase will resolve most of (3); the duplicated helpers in `server.py` still need to be deleted afterwards.

4. **Unrelated features mixed in** — `_heuristic_route`, `<system-reminder>` log filtering, response-content logging, fallback-reasons logging — all useful, none Anthropic-API-related. Belong in their own PRs.

The core feature (Anthropic Messages API passthrough via `/v1/messages`) is genuinely valuable and the conversion code in `anthropic_api.py` looks correct — the blockers are about scoping this back down to that feature and fixing the auth regression.

Plan (mirroring #30/#31):

- **By 2026-05-13** — if you can rebase, drop the unrelated work, fix the auth regression, and get the routing tests green, I'll re-review and merge.
- **After 2026-05-13** — I'll close as stale and carve the Anthropic-passthrough piece into a fresh PR with `Co-Authored-By` attribution to you, given the volume of overlap with already-merged work.

Let me know either way. Thanks for the original work — the core direction is right.

---

PR #37, opened 2026-04-18
feat: agent role detection for AI coding assistants

snowspig:
## Summary

- Add **agent role detection** — identify AI coding assistant session types (planning, explore, subagent) from system prompt markers
- **Phase-aware planning mode routing** — route differently based on the current planning phase:
  - **USER** (new request, no tool result) → reasoning model for decision-making
  - **EXPLORATION** (Read/Glob/Grep results) → fast model for speed
  - **PLAN_GENERATION** (Write/Edit tools) → reasoning model for quality
  - **CONTEXT** (indeterminate) → fast model as default
- **Explore agent** detection → routes to configurable `explore_model`
- **Subagent** detection → routes to configurable `subagent_model`
- Main session protection (>15KB system prompt) prevents misclassification

## New API

```python
# routing.py — new function
detect_agent_role(system_prompt, message_count, tool_names) -> {"role": str, "confidence": float, "signals": list}

# apply_routing_modifiers — new optional params
explore_model: Optional[str] = None
subagent_model: Optional[str] = None
```

## Motivation

When using NadirClaw as a proxy for AI coding assistants (Claude Code, Cursor, etc.), the system prompt contains markers that reveal the session type. Planning sessions should route differently based on the current phase — exploration can use fast/cheap models, while plan generation needs the best reasoning model. This provides significant cost savings without sacrificing quality.

## Testing

- 17 new unit tests: role detection, tool call extraction, planning phase routing
- All 79 existing routing tests still pass (no regressions)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #41, opened 2026-04-25
feat: show model success rate on dashboard cards

snowspig:
## Summary

Adds per-model success/failure rate tracking and display on the dashboard's model cards.

When models have failures (tracked via `fallback_reasons` in `requests.jsonl`), the model card now shows:
- Success count / total attempts (e.g. `48/50 (96%)`)
- Color coding: green (≥95%), yellow (≥80%), red (<80%)
- Categorized failure reasons: `rate_limit`, `disconnected`, `connection_error`, `timeout`, `other`

Models with no failures display unchanged — no noise for healthy models.

## Changes

- **`web_dashboard.py`**: Added `model_stats` to `/dashboard/api/stats` response — scans `fallback_reasons` to count per-model success/fail with reason classification. Updated model card rendering to show success rate when failures exist.

## Screenshot

Model cards now look like:
```
claude-sonnet-4-6
Requests  254
Success   232/254 (91%)
rate_limit:18, disconnected:4
Tokens    125,432
Cost      $1.2345
Avg latency 3200ms
```

## Test plan

- [ ] Run NadirClaw with multiple models and some fallbacks occurring
- [ ] Open `/dashboard` and verify model cards show success rate only for models with failures
- [ ] Verify models with 0 failures still show the original card layout
- [ ] Check color coding matches thresholds (green ≥95%, yellow ≥80%, red <80%)

### DIFF
--- a/nadirclaw/web_dashboard.py
+++ b/nadirclaw/web_dashboard.py
@@ -39,6 +39,47 @@ def _load_recent_logs(limit: int = 200) -> List[Dict[str, Any]]:
     return entries
 
 
+# Reason bucketing matches on error_type class names for stable classification.
+# Schema: fallback_reasons entries are {"model", "error_type", "message"} (PR #47).
+_REASON_BUCKETS: Dict[str, set] = {
+    "rate_limit": {"RateLimitExhausted", "RateLimitError"},
+    "timeout": {"ReadTimeout", "ConnectTimeout", "Timeout", "TimeoutError"},
+    "connection_error": {"ConnectError", "APIConnectionError"},
+    "disconnected": {"RemoteProtocolError", "EndOfStream"},
+}
+
+
+def _classify_error_type(error_type: str) -> str:
+    """Map an error_type class name to a dashboard reason bucket."""
+    for bucket, names in _REASON_BUCKETS.items():
+        if error_type in names:
+            return bucket
+    if error_type.endswith("ServerError"):
+        return "server_error"
+    return "other"
+
+
+def compute_model_stats(completions: List[Dict[str, Any]]) -> Dict[str, Dict[str, Any]]:
+    """Aggregate per-model success/fail counts and categorized reasons.
+
+    Success = request landed on `selected_model`. Fail = every entry in
+    `fallback_reasons`, attributed to the model that actually failed.
+    """
+    stats: Dict[str, Dict[str, Any]] = {}
+    for e in completions:
+        model = e.get("selected_model", "unknown")
+        stats.setdefault(model, {"success": 0, "fail": 0, "fail_reasons": {}})
+        stats[model]["success"] += 1
+        for fr in e.get("fallback_reasons") or []:
+            fm = fr.get("model", "unknown")
+            stats.setdefault(fm, {"success": 0, "fail": 0, "fail_reasons": {}})
+            stats[fm]["fail"] += 1
+            bucket = _classify_error_type(fr.get("error_type", ""))
+            reasons = stats[fm]["fail_reasons"]
+            reasons[bucket] = reasons.get(bucket, 0) + 1
+    return stats
+
+
 @router.get("/dashboard/api/stats")
 async def dashboard_stats(
     current_user: UserSession = Depends(validate_local_auth),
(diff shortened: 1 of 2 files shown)


doramirdor (commented):
Nice direction — surfacing per-model reliability on the dashboard is a real gap, and the color-coded thresholds are a clean UX choice. But there's one blocker and a few smaller things before this can land.

### 🚨 Blocking: `fallback_reasons` is never produced

The handler reads `e.get("fallback_reasons") or []` and counts every entry inside as a per-model failure with a reason. But that field doesn't exist on `main`:

```bash
$ grep -rn fallback_reasons nadirclaw/
# (no matches)
```

What `_log_request` *actually* writes today is a single string — the original failing model — under `fallback_used`:

```python
# nadirclaw/server.py:1389
"fallback_used": analysis_info.get("fallback_from"),
```

`_call_with_fallback` does accumulate the data you want (`failed_models` list at server.py:1011) and stamps `fallback_chain_tried` into `analysis_info` at line 1032, but the chain + the failure reason are never persisted to the log entry.

**Effect:** with this PR alone, every entry's `fallback_reasons or []` evaluates to `[]`, the per-model `fail` counter stays at 0 for every model, and the JS branch `if (fail > 0)` is never taken — the dashboard renders identically to today. Confirmed locally against `~/.nadirclaw/logs/requests.jsonl`: only `fallback_used` keys exist.

Two options:

**A. Land the producer first.** Modify `_call_with_fallback` (server.py:978) and `_stream_with_fallback` (server.py:1830) to record per-model failures with reason strings into `analysis_info["fallback_reasons"]`, then add a `"fallback_reasons": analysis_info.get("fallback_reasons", [])` line in the log entry at server.py:1389 and the streaming counterpart at server.py:1318. The data you need is right there in the `except` blocks at server.py:1037 and server.py:1908 — capture `type(chain_error).__name__` + `str(chain_error)[:200]` per failed model. This is the cleaner path; suggest splitting it into a small first PR (`feat: log per-model fallback reasons`) and rebasing this one on top.

**B. Use the existing `fallback_used` + `fallback_chain_tried` data.** If you don't want to expand the log schema, you can derive a partial picture from what's already logged: each completion has a single "this model failed first" via `fallback_used` (string). That gives you `fail += 1` for `fallback_used` per request, but no reason classification — drop the `fail_reasons` block in that case. Less rich, but works on data already in `requests.jsonl` / `requests.db` without a producer-side change.

I'd recommend (A) — the reason classification is the most useful part of the UX, and a no-op merge is worse than a held PR.

### ⚠️ Smaller things

**1. Reason classification — `disconnected` is too eager.** `"endofstream" in reason_raw.lower()` matches `EndOfStream` correctly, but `"connection" in reason_raw.lower()` matches `httpx.ConnectError`, `httpx.ConnectTimeout`, and any string containing the word "connection" in a free-text error message. Suggest using exception class names (you can stash `type(chain_error).__name__` directly during option A) and matching those instead of substring-on-error-text.

**2. `success` counter is per-completion's `selected_model`, which is the *post-fallback* model.** That's actually correct (haiku-after-opus-failed → haiku += 1 success, opus += 1 fail), but worth a comment so a future reader doesn't \"fix\" it.

**3. `mg.innerHTML += card` in a loop** rebuilds the DOM N times instead of once. Not a perf issue at this scale, but `parts.push(card); ... mg.innerHTML = parts.join('')` after the loop is the cheap form and matches the rest of the file's style.

**4. `name` is interpolated into `innerHTML` without escaping.** Pre-existing in this file (line 277 on main), so not a regression — but worth a follow-up to wrap with a small `escapeHTML()` helper since model names can come from user-edited config files.

**5. PR title vs branch name mismatch.** Branch is `feat/model-stats-and-domain-health` but the diff has no domain-health changes. Either drop the suffix from the branch name or split out the second feature into its own PR — keeps reviewability tight.

### CI

CI is green (3.10/3.11/3.12) but only because the unused-key path is a soft fallback. Once option (A) lands, please add a small unit test that posts a synthetic completion with `fallback_reasons` populated and asserts `model_stats[fm]["fail"] == 1` plus the reason classification. That's the regression test that protects this feature going forward.

Happy to re-review once the producer side is in place. Thanks for the contribution!

---

doramirdor:
@snowspig — friendly ping. This PR has been dormant since my review on 2026-04-25 (~10 days). The blocker hasn't moved: `fallback_reasons` is referenced in the dashboard handler but never produced anywhere in the codebase, so with this PR alone the success-rate panel renders identically to today (`fail` counter stays at 0 for every model).

Your two paths from the review still apply:

- **Option A (recommended)** — Land the producer first. ~10-line change in `_call_with_fallback` (server.py:978) and `_stream_with_fallback` (server.py:1830) to record `analysis_info["fallback_reasons"] = [{"model": ..., "reason": type(err).__name__, "detail": str(err)[:200]}]` from the existing `except` blocks at server.py:1037 / server.py:1908, plus a `"fallback_reasons": analysis_info.get("fallback_reasons", [])` line in the log entry. Suggest splitting that into a small first PR and rebasing this one on top.

- **Option B** — Use the existing `fallback_used` field (already logged today). You get per-request "which model failed first" → `fail += 1` for that model, but no reason classification. Simpler but less useful.

Plan, mirroring #30/#31/#39:

- **By 2026-05-13** — if you can land the producer side and rebase, I'll re-review and merge.
- **After 2026-05-13** — I'll close as stale or take it over with `Co-Authored-By` attribution.

Note for context: I just merged the rebase of your other approved PR (#37) as #45, since that one was sitting on the prior 2026-05-06 deadline. The model success rate UX is genuinely useful — just needs the data underneath. Let me know either way.

---

doramirdor (approved):
LGTM. Rebase on #47 is clean, schema match is correct (`model` / `error_type` / `message`), and the switch from substring matching to exact class-name buckets removes the `ConnectionPoolTimeoutError → connection_error` false positive I flagged.

The `compute_model_stats` extraction makes the aggregation directly testable, and the 14 new tests cover every bucket plus the reviewer's 48/50 scenario, missing fields, unknown error types, and explicit no-false-positive coverage for `ConnectionPoolTimeoutError`.

Buckets line up with the actual exception classes raised through `_call_with_fallback` / `_stream_with_fallback` (`RateLimitExhausted` from litellm, `ReadTimeout` / `ConnectTimeout` / `ConnectError` from httpx, `APIConnectionError` from openai). Unknown classes degrade to `other` gracefully.

Squash-merging. Thanks for the rebase!

---

PR #42, opened 2026-04-28
add update-models metadata refresh

ZardLi1115:
## Summary

Adds a `nadirclaw update-models` CLI command and local model metadata support so model context windows, pricing, and capabilities can be refreshed or overridden without editing `routing.py`.

## Changes

- Add `nadirclaw update-models` with `--output`, `--dry-run`, `--format json`, and optional `--source-url` / `NADIRCLAW_MODEL_REGISTRY_URL` support.
- Add `nadirclaw/model_metadata.py` for parsing, loading, and writing generated/user model metadata.
- Merge `~/.nadirclaw/models.json` and `~/.nadirclaw/models.local.json` into the runtime model registry.
- Add DeepSeek V4 Flash/Pro registry entries, defaults, aliases, and README examples.
- Document local metadata files and mark the roadmap item complete.

## Validation

- `/opt/anaconda3/bin/python -m pytest tests/test_routing.py tests/test_setup.py` — 147 passed, 1 warning
- `env PYTHONPYCACHEPREFIX=/tmp/nadirclaw-upload-pycache /opt/anaconda3/bin/python -m compileall nadirclaw tests` — passed

## Notes

Full test suite was attempted locally and still has unrelated environment failures around the Anaconda base install: `sentence_transformers/transformers/torch` version mismatch, missing `pytest-asyncio`, and local log directory permissions. The focused routing/setup tests covering this change pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to NadirClaw will be documented in this file.
 
+## [Unreleased]
+
+### Added
+- **`nadirclaw update-models` command** — writes refreshable model metadata to `~/.nadirclaw/models.json`, optionally merging a published registry JSON via `--source-url` or `NADIRCLAW_MODEL_REGISTRY_URL`.
+- **Local model metadata overrides** — the router now merges `~/.nadirclaw/models.json` and user-managed `~/.nadirclaw/models.local.json` into the runtime model registry.
+- **DeepSeek V4 explicit aliases** — added `deepseek-v4`, `deepseek-v4-flash`, and `deepseek-v4-pro` while preserving the existing `deepseek` alias for `deepseek/deepseek-chat`.
+
 ## [0.14.0] - 2026-04-03
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -635,7 +635,7 @@ Use short names instead of full model IDs:
 | `gpt5` | `gpt-5.2` |
 | `flash` | `gemini-2.5-flash` |
 | `gemini-pro` | `gemini-2.5-pro` |
-| `deepseek` | `deepseek/deepseek-v4-flash` |
+| `deepseek` | `deepseek/deepseek-chat` |
 | `deepseek-v4` | `deepseek/deepseek-v4-flash` |
 | `deepseek-v4-flash` | `deepseek/deepseek-v4-flash` |
 | `deepseek-v4-pro` | `deepseek/deepseek-v4-pro` |
@@ -740,6 +740,9 @@ Without options it exports the built-in registry. Pass `--source-url` or set
 `NADIRCLAW_MODEL_REGISTRY_URL` to merge a published registry JSON before saving. The
 router merges the saved file at startup, then applies any user-managed overrides from
 `~/.nadirclaw/models.local.json`.
+
+`update-models` only rewrites the generated metadata file. It does not re-export
+entries from `models.local.json`, so local overrides stay separate across refreshes.
 
 Use `models.local.json` for private models or custom pricing:
 
--- a/README.md
+++ b/README.md
@@ -283,6 +283,7 @@ NADIRCLAW_FALLBACK_CHAIN=gpt-4.1,claude-sonnet-4-5-20250929,gemini-2.5-flash  # 
 | **Claude + Claude** | `claude-haiku-4-5-20251001` | `claude-sonnet-4-5-20250929` | `ANTHROPIC_API_KEY` |
 | **OpenAI + Ollama** | `ollama/llama3.1:8b` | `gpt-4.1` | `OPENAI_API_KEY` |
 | **OpenAI + OpenAI** | `gpt-4.1-mini` | `gpt-4.1` | `OPENAI_API_KEY` |
+| **DeepSeek + DeepSeek** | `deepseek/deepseek-v4-flash` | `deepseek/deepseek-v4-pro` | `DEEPSEEK_API_KEY` |
 | **OpenAI Codex** | `gemini-2.5-flash` | `openai-codex/gpt-5.3-codex` | `GEMINI_API_KEY` + OAuth login |
 | **Fully local** | `ollama/llama3.1:8b` | `ollama/qwen3:32b` | None |
 
@@ -634,7 +635,10 @@ Use short names instead of full model IDs:
 | `gpt5` | `gpt-5.2` |
 | `flash` | `gemini-2.5-flash` |
 | `gemini-pro` | `gemini-2.5-pro` |
-| `deepseek` | `deepseek/deepseek-chat` |
+| `deepseek` | `deepseek/deepseek-v4-flash` |
+| `deepseek-v4` | `deepseek/deepseek-v4-flash` |
+| `deepseek-v4-flash` | `deepseek/deepseek-v4-flash` |
+| `deepseek-v4-pro` | `deepseek/deepseek-v4-pro` |
 | `deepseek-r1` | `deepseek/deepseek-reasoner` |
 | `llama` | `ollama/llama3.1:8b` |
 
@@ -693,6 +697,7 @@ If the estimated token count of a request exceeds a model's context window, Nadi
 nadirclaw setup              # Interactive setup wizard (providers, keys, models)
 nadirclaw serve              # Start the router server
 nadirclaw serve --log-raw    # Start with full request/response logging
+nadirclaw update-models      # Refresh local model metadata
 nadirclaw test               # Probe each configured model and verify it responds
 nadirclaw optimize <file>    # Test context compaction on a file (dry-run)
 nadirclaw classify <prompt>  # Classify a prompt (no server needed)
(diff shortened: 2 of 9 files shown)


doramirdor (commented):
Solid first contribution — ROADMAP item completed, clean separation between `BUILTIN_MODEL_REGISTRY` (frozen at import) and the merged runtime view, defensive `.get()` updates in `check_context_window` / `get_context_window` / `estimate_cost` so user-supplied incomplete entries don't crash the router, and the import-time merge failure path correctly logs and skips invalid files instead of taking the package down. Tests look good (`pytest tests/test_routing.py tests/test_setup.py` → 147 passed locally). A few things to address before merge — most are non-blocking polish, but one is a real breaking change that needs a release note.

### 🚨 Worth a release note

**`deepseek` alias quietly retargets** — `MODEL_ALIASES["deepseek"]` flips from `deepseek/deepseek-chat` ($0.28/$0.42, 128k) to `deepseek/deepseek-v4-flash` ($0.14/$0.28, 1M). Anyone using the `deepseek` short name in `--simple-model deepseek` or in `presets.yaml` lands on a different model with different pricing and context window. Cheaper *and* bigger window, so the change probably wins on substance — but it's a breaking change for explicit `deepseek` users and worth calling out in the PR description and CHANGELOG when this ships.

Two options if you want to keep this PR minimal: (a) leave `deepseek` → `deepseek/deepseek-chat` and only add the new explicit aliases (`deepseek-v4-flash`, `deepseek-v4-pro`), or (b) keep the retarget but add a one-liner to the README's "Aliases" section flagging the change. I'd take (a) — the principle of least surprise on a router that ships in production.

### ⚠️ Quick fixes

**1. Confirm the LiteLLM model IDs.** `deepseek/deepseek-v4-flash` and `deepseek/deepseek-v4-pro` need to match what LiteLLM accepts on the wire. If the upstream IDs are different (say `deepseek/v4-flash` or `deepseek-chat-v4`), users hitting these aliases get provider 404s rather than the price they expected from the registry. Have you smoke-tested an actual completion through `nadirclaw test` against both? If so, please drop the output in the PR description; if not, that's the one validation step that matters most for a model-metadata PR.

**2. `urllib.request.urlopen` accepts `file://` and `ftp://`.** For `--source-url` on a CLI flag, fine — operator chooses. For `NADIRCLAW_MODEL_REGISTRY_URL` (env var path), I'd restrict to `http(s)://` so an attacker who can set env vars on the host can't trivially exfiltrate via `file:///etc/...` even though the JSON parse step would fail. One-line guard in `update_models`:
```python
if source and not source.startswith(("http://", "https://")):
    raise click.ClickException(f"Source URL must use http(s): {source}")
```
The CliRunner test uses `source.as_uri()` (which is `file://`) — that test would need to switch to a tiny `http.server` fixture, or use the `--source-url` path directly without env var (since the flag is operator-controlled).

**3. Network failure crashes with a traceback.** If `update-models` is run with `--source-url https://unreachable/`, `urllib.request.urlopen` raises `URLError` and the CLI dies with a stack. Wrap in try/except and `raise click.ClickException(str(e))` — same shape as how Click handles user errors elsewhere.

**4. `write_model_metadata` is not atomic.** `path.write_text(...)` truncates and then writes; if the process is killed mid-write or two `update-models` runs race, `~/.nadirclaw/models.json` ends up half-written and the import-time merge logs a warning and skips it (so it's not catastrophic, just confusing). Standard fix:
```python
tmp = path.with_suffix(path.suffix + ".tmp")
tmp.write_text(...)
os.replace(tmp, path)
```

**5. No schema validation on values.** `parse_model_metadata` accepts any dict, so `{"context_window": "lots", "cost_per_m_input": -5}` silently merges in and the negative number then yields a negative `estimate_cost` downstream. A minimal type-check in `parse_model_metadata` would catch obvious typos at refresh time rather than at request time:
```python
for key in ("context_window",):
    if key in info and not isinstance(info[key], int):
        raise ValueError(f"{model_id}.{key} must be an integer")
for key in ("cost_per_m_input", "cost_per_m_output"):
    if key in info and (not isinstance(info[key], (int, float)) or info[key] < 0):
        raise ValueError(f"{model_id}.{key} must be a non-negative number")
```

### Nits

- `update_models` references `BUILTIN_MODEL_REGISTRY` directly — fine, but worth noting in the README docs section that a user's `models.local.json` is *not* re-exported by `update-models`. Today's behavior is correct (overrides stay in `models.local.json`), but the docs should call out the round-trip semantics so a user doesn't expect their overrides to land in the generated `models.json`.
- Tiny: `setup.py` flips `reasoning.deepseek` from `deepseek/deepseek-reasoner` to `deepseek/deepseek-v4-pro`. If `v4-pro` doesn't have the same chain-of-thought characteristics as `deepseek-reasoner`, this silently degrades the reasoning tier for DeepSeek-only setups. Worth a sentence in the PR explaining the choice.
- The newline added at end of `cli.py` is a clean fix — no notes there.

### Summary

Approving the **structure** of this PR — the metadata layering, the `update-models` command, the user-override file are all the right shape and a real improvement over hardcoded constants. The blockers above are about (a) being explicit about the breaking alias change and (b) making sure the new DeepSeek model IDs actually resolve through LiteLLM.

Happy to merge once (1) is confirmed and the alias change is either rolled back or documented in a CHANGELOG entry. (2)–(5) are nice-to-haves — pick what you want to land here vs. follow-up.

---

doramirdor:
@ZardLi1115 — friendly heads-up: I drafted #43 as a follow-up that addresses the review feedback on this PR (CLI scheme parity, 10 MiB payload cap, narrower exception handler in `_merge_external_model_metadata`, plus updated tests covering scheme rejection / size cap / `models.local.json` precedence / the malformed-metadata warning path). It's a draft right now — CI is green on 3.10/3.11/3.12.

The breaking-change question on the `deepseek` alias retarget is still the main thing I'd like your take on before either of these merges:

- **Option A** — keep `deepseek` → `deepseek/deepseek-chat`, only add `deepseek-v4-flash` / `deepseek-v4-pro` as new explicit aliases (no surprise for existing `deepseek` users).
- **Option B** — keep the retarget but call it out in `CHANGELOG.md` as a breaking change.

I'd take (A) — least-surprise wins for a router people use in `presets.yaml`. Happy to update #43 either way once you weigh in. Also still need confirmation that `deepseek/deepseek-v4-flash` / `-v4-pro` resolve through LiteLLM on the wire (the one validation step that matters most for a model-metadata PR — `nadirclaw test` output would be enough).

If you'd rather pull #43's commits into your branch, that's totally fine — your name stays on the feature. Otherwise we can merge #43 with `Co-Authored-By` attribution to you. Let me know which you prefer.

ZardLi1115:
Thanks for the heads-up — Option A sounds right to me.

To be explicit, I agree we should preserve the existing aliases:

`deepseek` -> `deepseek/deepseek-chat`
`deepseek-r1` -> `deepseek/deepseek-reasoner`

And add the V4 models only as new explicit aliases:

`deepseek-v4` -> `deepseek/deepseek-v4-flash`
`deepseek-v4-flash` -> `deepseek/deepseek-v4-flash`
`deepseek-v4-pro` -> `deepseek/deepseek-v4-pro`

I also smoke-tested the new LiteLLM IDs with real DeepSeek calls using very
small requests (`max_tokens=1`). Both resolved successfully:
deepseek/deepseek-v4-flash: OK
usage: prompt_tokens=8, completion_tokens=1, total_tokens=9

deepseek/deepseek-v4-pro: OK
usage: prompt_tokens=8, completion_tokens=1, total_tokens=9

Feel free to pull #43’s commits into my branch if that’s easiest. I’m also
fine with merging #43 directly with Co-Authored-By attribution.

doramirdor (approved):
Re-reviewed against the 2026-04-28 feedback. All blockers addressed:

- ✅ **Alias backwards compat** — Option A landed cleanly. `deepseek` and `deepseek-r1` still resolve to chat/reasoner; V4 models are new explicit aliases (`deepseek-v4`, `deepseek-v4-flash`, `deepseek-v4-pro`).
- ✅ **LiteLLM resolution confirmed** — your `max_tokens=1` smoke calls are exactly the validation I needed. Both V4 IDs route on the wire.
- ✅ **CHANGELOG entry** — covers the command, the local-override file, and explicitly notes the existing `deepseek` alias is preserved.
- ✅ **Schema validation** — `_validate_model_info` rejects negative costs, non-int context windows, non-bool `has_vision`. The strict `type(...) is not int` (rather than `isinstance`) is the right call here — it correctly rejects `bool` values that would otherwise sneak through as ints.
- ✅ **Atomic write** — tmp + `os.replace` pattern is correct.
- ✅ **Env-var URL scheme check** — `http(s)` only, with a test covering the `file://` rejection path.
- ✅ **Test coverage** — 12+ new tests including dry-run, source-url merge, scheme rejection, network-failure path, and schema-rejection assertions. Nice work.

Three nice-to-haves remain — `--source-url` CLI flag scheme check, 10 MiB payload cap, and narrower `_merge_external_model_metadata` exception handler — all of which are already covered in #43, so I'll land that right after this. No need to do anything more on this branch.

Approving and merging.

---

PR #44, opened 2026-05-03
Add LLM cost change analysis to CI

Jwrede:
## Summary

Add [tokentoll](https://github.com/Jwrede/tokentoll) GitHub Action to analyze LLM API cost changes on pull requests.

When a PR modifies code that makes LLM API calls (model swaps, max_tokens changes, new call sites), tokentoll posts a comment showing the projected cost impact. It stays silent on PRs that don't touch LLM code.

## What it detects

tokentoll uses Python AST analysis (zero runtime dependencies) to find calls to:
- OpenAI (`chat.completions.create`, `responses.create`, `embeddings.create`)
- Anthropic (`messages.create`)
- Google GenAI (`generate_content`)
- LiteLLM (`completion`, `embedding`)
- LangChain (`ChatOpenAI`, `ChatAnthropic`, `init_chat_model`)
- Zhipu AI / GLM (`ZhipuAiClient`, `ZhipuAI`)

## tokentoll Scan Report

| File | Line | SDK | Model | Est. Cost/Call | Monthly |
|------|------|-----|-------|---------------|---------|
| `nadirclaw/cli.py` | 1409 | litellm | gpt-4o (default) | $0.001350 | $1.35 |
| `nadirclaw/server.py` | 621 | google_genai | gemini-2.0-flash (default) | $0.000869 | $0.87 |
| `nadirclaw/server.py` | 1776 | google_genai | gemini-2.0-flash (default) | $0.000869 | $0.87 |
| `nadirclaw/server.py` | 898 | litellm | gpt-4o (default) | $0.04 | $42.21 |
| `nadirclaw/server.py` | 1672 | litellm | gpt-4o (default) | $0.04 | $42.21 |

**Total estimated monthly cost: $87.51**

<details>
<summary>Assumptions</summary>

- 1000 calls/month per call site

</details>

*Generated by [tokentoll](https://github.com/Jwrede/tokentoll)*

## How it works

- Runs only on PRs, compares base vs head
- Posts a sticky comment (updates in place, no duplicates)
- Only comments when LLM calls are added, removed, or modified
- First run posts an initial scan (this PR) so you can see the output
- Zero runtime dependencies, ~1s scan time for most projects
- [PyPI](https://pypi.org/project/tokentoll/) | [GitHub](https://github.com/Jwrede/tokentoll)

## Changes

- Adds `.github/workflows/llm-costs.yml` (17 lines)
- No code changes, no new dependencies

### DIFF
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -5,6 +5,7 @@ on:
     branches: [main, master]
 
 permissions:
+  contents: read
   pull-requests: write
 
 jobs:
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -15,4 +15,4 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: Jwrede/tokentoll@v0.3.0
+      - uses: Jwrede/tokentoll@v0.4.0
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -1,0 +1,17 @@
+name: LLM Cost Analysis
+
+on:
+  pull_request:
+    branches: [main, master]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  cost-analysis:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: Jwrede/tokentoll@v0.3.0
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -15,4 +15,4 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: Jwrede/tokentoll@v0.4.0
+      - uses: Jwrede/tokentoll@a4c8ce5f97e751f34498f6503087e5583c02cbbc
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -15,4 +15,4 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: Jwrede/tokentoll@a4c8ce5f97e751f34498f6503087e5583c02cbbc
+      - uses: Jwrede/tokentoll@cfc9307c2e820de1223fc707ee7e945f6f972379  # v0.5.2
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -15,4 +15,4 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: Jwrede/tokentoll@ff50a35604b7858d9064cca6354b6229a94f7235  # v0.6.0
+      - uses: Jwrede/tokentoll@753ca4d1150c74169b52a843439049b65e256d2b  # v0.6.1
--- a/.github/workflows/llm-costs.yml
+++ b/.github/workflows/llm-costs.yml
@@ -15,4 +15,4 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: Jwrede/tokentoll@cfc9307c2e820de1223fc707ee7e945f6f972379  # v0.5.2
+      - uses: Jwrede/tokentoll@ff50a35604b7858d9064cca6354b6229a94f7235  # v0.6.0
--- a/.tokentoll.yml
+++ b/.tokentoll.yml
@@ -1,0 +1,1 @@
+default_model: gpt-4o
--- a/.tokentoll.yml
+++ b/.tokentoll.yml
@@ -1,1 +1,0 @@
-default_model: gpt-4o


doramirdor:
Thanks for the contribution and for being upfront that this is your own action — appreciated.

I want to be honest about why I'm hesitant to merge as-is:

### 🚨 Signal-to-noise on *this* repo

Your own scan output (in the PR description) shows the issue: every flagged call site resolves to `<dynamic>` model / `N/A` cost / `$0.000000` monthly. That's not a false positive — it's the architecture. NadirClaw is a *router*: model names are decided at runtime by the classifier and config, never hardcoded. The five flagged sites are all `litellm.acompletion(**call_kwargs)` / `genai.Client(...).models.generate_content(model=settings.SOME_MODEL, ...)` patterns where the model arg is built from `Settings`/routing state, not a string literal.

So the steady-state behavior here would be: tokentoll posts a sticky comment on every PR saying "no LLM cost change detected" because there's nothing static for it to compare. That's noise, not signal. The tool is built for app code that calls `client.chat.completions.create(model="gpt-4o", ...)` directly — a router is the wrong shape for it.

### 🔒 Supply-chain pinning

```yaml
- uses: Jwrede/tokentoll@v0.3.0
```

Pinning to a mutable tag rather than a commit SHA means a force-pushed `v0.3.0` would silently land in CI with `pull-requests: write`. GitHub's [own guidance](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) is to pin third-party actions to a full SHA. Not a blocker on its own, but combined with the signal issue it tips the balance.

### Path forward

Two things that would make me reconsider:

1. **A dynamic-model mode** — for routers/proxies, surfacing changes like "this PR added a new tier", "this PR changed the default classifier threshold", or "this PR added a new provider integration" would be the equivalent useful signal. That's harder, but it's the version that helps this class of repo.
2. **SHA pinning** in the workflow file.

Leaving open for now so you can decide whether either of those is in scope. As written, I'd lean toward not merging — a workflow that produces empty comments forever isn't a fair deal for either side. Thanks again for thinking of the project.

Jwrede:
Thanks for the thoughtful feedback, you're right that the initial scan was all `<dynamic> | N/A` which isn't useful for a router.

I just shipped v0.4.0 with a `.tokentoll.yml` config file that addresses this:

```yaml
default_model: gpt-4o
```

When tokentoll can't statically resolve a model name, it falls back to the configured default for cost estimation. The output now shows `gpt-4o (default)` with real cost numbers instead of `N/A`. The config also supports per-path overrides if different parts of the codebase use different model tiers.

This PR now includes a `.tokentoll.yml` with `gpt-4o` as the default, you can adjust it to whatever model best represents your typical traffic. The action is also bumped to v0.4.0.

Re: SHA pinning, fair point, I'll look into pinning to a commit SHA in the workflow.

Jwrede:
Update: both concerns are now addressed.

**SHA pinning.** The workflow now pins to a full commit SHA:

```yaml
- uses: Jwrede/tokentoll@753ca4d1150c74169b52a843439049b65e256d2b  # v0.6.1
```

**Dynamic model scan output.** Since v0.4.0 (and now v0.6.1 which is what this PR pins), tokentoll applies per-SDK defaults when a model name cannot be resolved statically. The refreshed scan in the PR description above shows what this looks like for NadirClaw:

| File | SDK | Model | Monthly |
|------|-----|-------|---------|
| cli.py:1409 | litellm | gpt-4o (default) | $1.35 |
| server.py:621 | google_genai | gemini-2.0-flash (default) | $0.87 |
| server.py:898 | litellm | gpt-4o (default) | $42.21 |
| ... | | | |
| **Total** | | | **$87.51** |

The `(default)` label makes it explicit that the model is a guess rather than a resolved name. If NadirClaw's typical traffic is dominated by a specific tier (e.g. `gemini-2.0-flash` for all litellm calls too), you can override per-SDK defaults in a `.tokentoll.yml`:

```yaml
default_models:
  litellm: gemini-2.0-flash
```

If you'd rather suppress cost estimation for all dynamic calls entirely, v0.6.1 adds an opt-out:

```yaml
skip_dynamic_models: true
```

The steady-state behavior with defaults enabled: tokentoll only comments when a PR adds, removes, or changes an LLM call site. If a PR touches none, there's no comment. It would comment on changes like "added a new litellm.acompletion call site in a new handler" or "removed one of the existing call sites," which seems useful even for a router.

Let me know if this looks reasonable or if you'd still prefer not to merge.

doramirdor:
Thanks @Jwrede — the SHA pin and the per-SDK defaults both address what I raised, and the v0.6.1 `skip_dynamic_models` opt-out is a thoughtful escape hatch. I appreciate the iteration speed.

I'm still going to lean toward not merging, but the reason has shifted from "the tool produces no useful output here" to a different concern, so I want to be clear about what's changed:

**With defaults enabled** — every PR that touches a router internal will get a comment showing a $87/mo total built from `gpt-4o (default)` and `gemini-2.0-flash (default)` rows. None of those numbers reflect what NadirClaw users actually pay (their bill is determined at routing time by *their* configured tiers — for many users on a default config, that's `gemini-3-flash-preview` + `openai-codex/gpt-5.3-codex`, not gpt-4o). The risk isn't that the comment is empty — it's that it'll be confidently wrong, and on a routing project of all places, "your changes cost $42/mo" lands as a precise-but-incorrect signal that I'd have to explain in every PR review.

**With `skip_dynamic_models: true`** — we get the change-detection signal ("a new litellm call site appeared in `server.py`") without the misleading dollar figures. That's the version of the tool I'd actually consider merging, but at that point what we're getting is essentially a `grep` for new call sites, which is a thinner value proposition than the project's positioning suggests.

**The honest answer** is that NadirClaw is just a hard fit for tokentoll's strengths — the same architecture that makes the tool produce `(default)` rows here is what the project exists to be. That's not a tooling failure; static analysis of LLM costs and a runtime LLM router are solving inverse problems.

I'll close in a few days unless you want the `skip_dynamic_models: true` variant landed — happy to merge that version if you're game. Either way, real appreciation for the contribution and for the patient back-and-forth. Best of luck with tokentoll; I think the tool's a good fit for app code that calls providers directly.

PR #47, opened 2026-05-07
feat: log per-model fallback reasons

ZardLi1115:
## Summary

Adds producer-side logging for per-model fallback reasons.

This records an ordered `fallback_reasons` list when fallback candidates fail, so the dashboard work in #41 and the follow-up provider-health work can share the same failure signal.

Each fallback reason includes:

- `model`
- `error_type = type(err).__name__`
- `message = str(err)[:200].replace("\n", " ")`

## Scope

- Record non-streaming primary fallback failure reasons
- Record non-streaming fallback-chain failure reasons
- Record streaming pre-content fallback failure reasons
- Skip mid-stream failures in this PR
- Include `fallback_reasons` in request log entries
- Persist `fallback_reasons` to SQLite as JSON text
- Add focused tests for fallback reason shape and SQLite persistence

## Non-goals

- No provider-health routing behavior
- No dashboard UI changes
- No mid-stream interruption schema

## Testing

- `/opt/anaconda3/envs/myenv/bin/python -m pytest tests/test_fallback_chain.py tests/test_request_logger.py -q`
- `/opt/anaconda3/envs/myenv/bin/python -m pytest -q`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to NadirClaw will be documented in this file.
 - **`nadirclaw update-models` command** — writes refreshable model metadata to `~/.nadirclaw/models.json`, optionally merging a published registry JSON via `--source-url` or `NADIRCLAW_MODEL_REGISTRY_URL`.
 - **Local model metadata overrides** — the router now merges `~/.nadirclaw/models.json` and user-managed `~/.nadirclaw/models.local.json` into the runtime model registry.
 - **DeepSeek V4 explicit aliases** — added `deepseek-v4`, `deepseek-v4-flash`, and `deepseek-v4-pro` while preserving the existing `deepseek` alias for `deepseek/deepseek-chat`.
+- **Fallback reasons logging** — failed fallback attempts now record ordered per-model `fallback_reasons` with compact error types and sanitized messages.
 
 ## [0.14.0] - 2026-04-03
 
--- a/nadirclaw/request_logger.py
+++ b/nadirclaw/request_logger.py
@@ -64,6 +64,7 @@ def _init_db() -> None:
                     daily_spend REAL,
                     response_preview TEXT,
                     fallback_used TEXT,
+                    fallback_reasons TEXT,
                     error TEXT,
                     tool_count INTEGER,
                     has_images INTEGER,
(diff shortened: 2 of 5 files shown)


doramirdor (approved):
## Code Review

LGTM. This matches the spec from #46 exactly and the implementation is clean.

### What I checked

**Spec conformance** — every bullet from the issue thread is implemented:
- `error_type = type(err).__name__` ✓ (`_fallback_reason` helper at server.py:33)
- `message = str(err)[:200].replace("\n", " ")` ✓
- Pre-content failure capture only for streaming ✓ (mid-stream correctly skipped, server.py:2001)
- Producer locations match the spots from the issue thread:
  - Non-streaming primary failure — server.py:1018
  - Non-streaming chain failure — server.py:1050
  - Streaming pre-content failure — server.py:2001
- Log entries include `fallback_reasons` for both paths (server.py:1370 streaming, server.py:1442 non-streaming)
- `analysis_info.setdefault("fallback_reasons", []).append(...)` preserves insertion order ✓
- `dict(analysis_info)` shallow-copy at server.py:1331 still shares the inner list reference path correctly — verified the streaming wrapper's `_stream_analysis.get("fallback_reasons", [])` sees what `_stream_with_fallback` accumulated.

**Storage** — SQLite schema migration via `ALTER TABLE ADD COLUMN fallback_reasons TEXT` follows the existing additive-column pattern. JSON-serialized into TEXT column. Empty list → `None` stored, which is the right call (no need to round-trip empty arrays).

**Tests** — `test_fallback_chain.py` asserts the shape on the primary→backup path, `test_request_logger.py` round-trips through SQLite with `json.loads`. Coverage is appropriate for the scope.

**CI** — green on 3.10/3.11/3.12.

### Nits (non-blocking, fine to defer)

1. **`setdefault().append()` repeated 3 times** — could pull into a small helper like `_record_fallback_reason(analysis_info, model, error)` to centralize the append. Minor; readable as-is.
2. **Test coverage on the chain-cascade and all-fail paths** — the cascade test at `test_fallback_cascade_through_chain` would benefit from one additional assertion that `fallback_reasons` has length == failure count. Worth a one-line follow-up but not blocking.
3. **CHANGELOG** — entry is good; consider tagging this for the next minor release (0.15.0) since it's a log-schema addition that downstream tooling (#41) reads.

### Downstream impact

This unblocks both #41 (dashboard reads `fallback_reasons` from log entries) and #46 (provider-health tracker reads the in-memory signal before it hits the log). Three pieces of work, one shared data shape — exactly what we wanted.

Approving. Will merge after a final read-through, or feel free to push if you spot anything before then.

---

PR #48, opened 2026-05-08
feat: add provider health-aware fallback routing

ZardLi1115:
## Summary
- add opt-in in-process provider health tracking behind NADIRCLAW_PROVIDER_HEALTH=true
- prefer healthy fallback candidates while keeping cooling-down models as a last resort
- expose /internal/provider_health only when provider health routing is enabled

## Scope
- primary model selection is unchanged; health ordering only applies to fallback candidates
- RateLimitExhausted / 429 is recorded but does not trip the unhealthy bit in v1
- timeouts, connection errors, and provider 5xx-style errors count as health failures
- health state is in-memory only and is not persisted across restarts

## Testing
- /opt/anaconda3/envs/myenv/bin/python -m pytest tests/test_provider_health.py tests/test_fallback_chain.py tests/test_server.py tests/test_streaming_fallback.py -q
- /opt/anaconda3/envs/myenv/bin/python -m pytest -q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to NadirClaw will be documented in this file.
 - **Local model metadata overrides** — the router now merges `~/.nadirclaw/models.json` and user-managed `~/.nadirclaw/models.local.json` into the runtime model registry.
 - **DeepSeek V4 explicit aliases** — added `deepseek-v4`, `deepseek-v4-flash`, and `deepseek-v4-pro` while preserving the existing `deepseek` alias for `deepseek/deepseek-chat`.
 - **Fallback reasons logging** — failed fallback attempts now record ordered per-model `fallback_reasons` with compact error types and sanitized messages.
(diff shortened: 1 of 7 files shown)


doramirdor (approved):
LGTM — exact scope we discussed in #46, all six v1 invariants honored.

Verified locally on `d332f67`:
- `pytest` → 568 passed (full suite, all 3 Python matrix versions green in CI too)
- `record_failure("RateLimitExhausted")` correctly does **not** trip the health bit
- `record_success` resets cooldown
- `_record_provider_*` is called for primary too (correct — primary _selection_ is unaffected, but we want primary failures to inform _future_ fallback ordering when this model appears in someone else's chain)
- `/internal/provider_health` returns 404 when flag is off, snapshot when on

One small nit (non-blocking, can land in a follow-up if you want):

`_provider_health_tracker()` mutates `failure_threshold` / `cooldown_seconds` on the singleton on every call. Works fine since env vars don't change at runtime, but it's a minor code smell. Cleaner long-term shape would be to read these as `@property` on the tracker (resolved at use time) or pass them through to `is_available` / `record_failure`. Not a merge blocker — happy to accept as-is.

Squashing in.

---

PR #84, opened 2026-08-11
fix(server): reconcile request params the routed model rejects (#83)

matteoantoci:
Fixes #83.

## Problem

Clients pick request parameters from the model id **they** can see. Behind this proxy that id is a router alias (`nadir-auto`, `nadir-eco`), which no capability table knows, so the client assumes the newest model capabilities. Routing then selects the real model **after** that decision. When routing picks an older, cheaper model, the request carries parameters that model rejects, and every request 400s.

The cheap tier is therefore unusable under Claude Code: the moment the classifier says `simple`, the request fails. That defeats the point of the router.

I originally reported this as "the client always sends adaptive thinking". That was wrong, and I corrected it [in the issue](https://github.com/NadirRouter/NadirClaw/issues/83#issuecomment-5252363588): the client does check the model, it just cannot check a model it has not been told about. This is why the fix belongs here — the proxy is the only component that knows both the requested parameters and the model that will serve them.

Measured before this change, `claude -p --model nadir-eco` with `NADIRCLAW_SIMPLE_MODEL=claude-haiku-4-5-20251001`, against a live Anthropic subscription:

```
nadir-eco -> claude-haiku-4-5-20251001  error  "adaptive thinking is not supported on this model"
nadir-eco -> claude-haiku-4-5-20251001  error  "This model does not support the effort parameter."
nadir-eco -> claude-haiku-4-5-20251001  error  "role 'system' is not supported on this model"
nadir-eco -> claude-haiku-4-5-20251001  ok
```

Three failed upstream calls per request, repeated identically on every request. The user still got an answer only because Claude Code degrades its own request shape and retries.

## What this does

`/v1/messages` reconciles the request against the 400 it gets back and retries **within the same request**, bounded to 4 attempts:

| Anthropic 400 | Fix | Modifier recorded |
|---|---|---|
| `adaptive thinking is not supported` | `thinking: {"type":"adaptive"}` → `{"type":"enabled","budget_tokens":N}`, clamped to `1024 <= N < max_tokens` and capped at 32000 | `thinking_downgrade(adaptive→enabled, budget=N)` |
| `does not support the effort parameter` | drop `output_config.effort` (or top-level `effort`) | `effort_drop(high)` |
| `role 'system' is not supported` | fold `system`/`developer` turns into the top-level `system` field | `system_role_fold(N)` |

Each fix is remembered per model in a process-local cache and pre-applied on later requests, so discovery costs one round trip per model per process, not one per request. An unrecognised 400 is returned untouched and never retried.

Applied fixes land in `requests.jsonl` as `modifiers_applied` entries, alongside the existing `context_window_swap` and `agent_role` modifiers, and in an `X-NadirClaw-Params-Reconciled` response header. The existing `max_tokens` clamp (#73) now runs inside the same loop and keeps its own header.

## Why this shape

- **It mirrors a pattern already in the codebase.** The `max_tokens` clamp-retry for #73 does exactly this: read the ceiling out of the 400, fix the body, retry once. This generalises that loop rather than adding a second mechanism.
- **No capability list to maintain.** Hardcoding which models support which parameter would need syncing on every model release, and it fails in the wrong direction: an unknown new model would be assumed capable and 400. Learning from the error stays correct as the API changes.
- **`thinking` intent is preserved, not discarded.** Older models accept `enabled` with a budget, so the request still thinks. Only `effort`, which merely shapes how hard the model tries, is dropped.
- **The system fold already exists here.** 0.21.1 applied the same fix to the OAuth completion path for #74. This brings the passthrough in line.
- **Bounded.** 4 attempts covers the three known fixes plus the clamp; a misbehaving upstream cannot loop.

## Evidence after

Same command, same config, after the change:

```
nadir-eco -> claude-haiku-4-5-20251001  ok  in=1077 out=43  cost=$0.001292
  modifiers_applied=[thinking_downgrade(adaptive→enabled, budget=31999),
                     effort_drop(high), system_role_fold(1)]
```

One upstream call, no errors. Two further requests produced one call each and **zero** new reconcile warnings in the server log, which confirms the cache: the three `rejected a parameter` warnings appear exactly once per model per process.

## Tests

16 new tests in `tests/test_server.py::TestParamReconcile` cover each fixer in isolation (including the budget clamp, the no-room-for-a-budget case, and the fold that would leave zero messages), the error matching, the per-model cache, the log modifier, and four end-to-end cases through the endpoint: convergence over successive 400s, a second request with no wasted call, an unknown 400 passed through unretried, and the streaming path retrying before any SSE byte is forwarded.

`pytest` on a clean checkout: **886 passed, 1 skipped**.

One note in case you see it: with a populated `~/.nadirclaw/.env` I get 10 pre-existing failures in `test_routing.py`/`test_server.py`, because those tests assert on default model names and the settings module reads the user's real `.env`. They fail identically on `main` without this branch, and all pass with a pristine `HOME`. Unrelated to this change, but it might be worth isolating settings in those tests.

## Not included

The OpenAI-compatible `/v1/chat/completions` path forwards `thinking` at three sites and can hit the same 400 for OpenClaw and Codex users. The helpers here are path-agnostic, so wiring them in there is small, but I kept this PR to the path where the bug reproduces. Happy to extend it in this PR if you prefer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable changes to NadirClaw will be documented in this file.
 ### Changed
 - **License changed from MIT to the PolyForm Noncommercial License 1.0.0.** NadirClaw is now free for any noncommercial use (personal, research, education, evaluation, and noncommercial organizations). Commercial use requires a separate commercial license, available via [getnadir.com](https://getnadir.com). This applies to new versions; releases previously published under MIT remain available under MIT. The bundled `wide_deep_asym_v3` classifier weights and the `cascade-verifier-v1` snapshot are now released under the same noncommercial terms.
 
+### Fixed
+- **`/v1/messages` forwarded parameters the routed model cannot accept, so cheap tiers 400 on every request** — clients pick request parameters from the model id they can see, which behind the router is an alias (`nadir-auto`), so they assume the newest capabilities. When routing then selected an older, cheaper model, the request carried parameters that model rejects and every request failed (#83). The passthrough now reconciles the request against the 400 it gets back and retries within the same request, bounded to 4 attempts: `thinking: {"type":"adaptive"}` is rewritten to `{"type":"enabled","budget_tokens":N}`, an unsupported `effort` hint is dropped, and `system`/`developer` turns are folded into the top-level `system` field (the same fix already applied to the OAuth completion path in 0.21.1). Each fix is remembered per model, so the discovery costs one round trip per model per process instead of one per request. Applied fixes are recorded in `requests.jsonl` as `modifiers_applied` entries and signalled by an `X-NadirClaw-Params-Reconciled` response header. Verified against a live Anthropic subscription: `claude-haiku-4-5` went from 3 failed upstream calls per Claude Code request to a single successful one.
+
 ## [0.21.1] - 2026-06-25
 
 ### Added
--- a/nadirclaw/server.py
+++ b/nadirclaw/server.py
@@ -2484,6 +2484,21 @@ _ANTHROPIC_COUNT_TOKENS_UPSTREAM = "https://api.anthropic.com/v1/messages/count_
 # Anthropic 400 when the requested max_tokens exceeds the routed model's ceiling:
 #   "max_tokens: 100000 > 64000, which is the maximum allowed ..."
 _MAX_TOKENS_400_RE = re.compile(r"max_tokens:\s*(\d+)\s*>\s*(\d+)")
+# Anthropic 400s raised when a client sends a parameter the routed model does not
+# support. Clients pick these parameters from the model id they can see, which
+# behind this proxy is a router alias, so they assume the newest capabilities (#83).
+_ADAPTIVE_THINKING_400_RE = re.compile(r"adaptive thinking is not supported", re.I)
+_EFFORT_400_RE = re.compile(r"does not support the effort parameter", re.I)
+_SYSTEM_ROLE_400_RE = re.compile(r"role '?system'? is not supported", re.I)
+# Anthropic requires 1024 <= budget_tokens < max_tokens for `thinking.type=enabled`.
+_MIN_THINKING_BUDGET = 1024
+_MAX_THINKING_BUDGET = 32000
+# Upper bound on reconcile round trips for one request, so a misbehaving upstream
+# can never loop. Four covers the known fixes plus the max_tokens clamp (#73).
+_MAX_RECONCILE_ATTEMPTS = 4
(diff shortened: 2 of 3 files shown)


doramirdor (approved):
Thank you for this — the follow-up correction on #83 in particular. Narrowing the mechanism from "the client always sends adaptive" to "the client cannot check a model it has not been told about" is what makes the fix belong here rather than in a capability list, and I agree with that reasoning: a list would need syncing on every model release and would fail open, 400ing on exactly the models we know least about.

I checked the branch out and ran the whole suite on a pristine `HOME`: **886 passed, 1 skipped**, matching your numbers. Reviewed the integration points too — in particular `_preapply_known_fixes` running before `_inject_claude_code_identity` is correct, because the injector already normalises `system` from both the string and the block-array form, so a fold that has rewritten `system` into blocks is handled.

Approving. Three notes, none blocking.

**1. The non-streaming loop applies a fix on its last attempt and then discards it.**

`for _attempt in range(_MAX_RECONCILE_ATTEMPTS)` — on the final iteration a 400 still runs the fixers, appends to `reconciled`, and then the loop ends without re-posting. So `requests.jsonl` can record a `modifiers_applied` entry for a fix that never went on the wire, which is the one thing the modifier log is supposed to make impossible. The streaming path already has the right guard (`attempt < _MAX_RECONCILE_ATTEMPTS - 1`); mirroring it in the non-streaming loop would make the two paths agree.

**2. A matched-but-unapplied reconciler still gets cached.**

`_reconcile_from_error` does `_MODEL_PARAM_FIXES.setdefault(model, set()).add(name)` before calling the fixer, so when a fixer declines — a body of nothing but system turns, which `_fold_system_messages` correctly refuses to fold — the model is recorded as needing a fix it never received. Harmless in effect, since pre-apply is then a no-op and the empty `fixes` list still breaks the loop rather than spinning, but the cache ends up describing something that did not happen. Recording the name only when the fixer returns a modifier would keep it honest.

**3. Unrelated to you, but worth saying here:** CI runs `pytest tests/ -v --ignore=tests/test_server.py`, so the 289 lines of tests you added never execute on GitHub. I ran them locally for this review and I am fixing the exclusion separately — not something for this PR.

On your two other observations: the `~/.nadirclaw/.env` bleed into `test_routing.py`/`test_server.py` is real and pre-existing, and yes please on extending this to the OpenAI-compatible path. Keep this PR to `/v1/messages` where the bug reproduces; I will follow up on `/v1/chat/completions` (it forwards `thinking` at three sites, two of them through LiteLLM, so the retry there needs a different hook than a raw `httpx` status check).

